PR #111, opened 2020-12-29
fix(notification-service): update method to fetch user id for notification user

mayank-SFIN571:
## Description

update method to fetch user id for notification user.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

### DIFF
--- a/services/notification-service/src/controllers/notification.controller.ts
+++ b/services/notification-service/src/controllers/notification.controller.ts
@@ -11,23 +11,24 @@ import {
   getFilterSchemaFor,
   getModelSchemaRef,
   getWhereSchemaFor,
+  HttpErrors,
   param,
   post,
   requestBody,
-  HttpErrors,
 } from '@loopback/rest';
 import {CONTENT_TYPE, STATUS_CODE} from '@sourceloop/core';
-import {authenticate, STRATEGY, AuthErrorKeys} from 'loopback4-authentication';
+import {authenticate, AuthErrorKeys, STRATEGY} from 'loopback4-authentication';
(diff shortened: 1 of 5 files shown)


services/notification-service/src/controllers/notification.controller.ts
```diff
@@ -188,7 +191,7 @@ export class NotificationController {
     return notif.receiver.to.map(to => {
       const notifUser = new NotificationUser();
       notifUser.notificationId = notif.id ?? '';
-      notifUser.userId = to.id;
+      notifUser.userId = this.notifUserService.getUserId(to);
```
samarpan-b, line 194: lets move this out of loop. and make this async because this may contain DB hits further. it should take receiver as parameter and return a promise.

samarpan-b (changes requested):

services/notification-service/src/controllers/notification.controller.ts
mayank-SFIN571: I am planning to send an extra key to the 'receiver.to' which will be used as 'userId'.

services/notification-service/src/types.ts
```diff
@@ -8,6 +9,10 @@ export interface IChannelManager {
   ): boolean;
 }
 
+export interface INotificationUserManager {
+  getNotifUsers(notif: Notification): NotificationUser[];
```
samarpan-b, line 13: it should return Promise<NotificationUser[]>

samarpan-b (changes requested):

samarpan-b (approved):

Issue #122, opened 2021-03-06
/notification-users get api is not working for userTenantId (issue of @sourceloop/notification-service)


sandeep-sourcefuse:
**Is your feature request related to a problem? Please describe.**
/notification-users get api is not working for userTenantId because we are using user tenant id for our socket connection which is **"to"** parameter in request and currently it is comparing userId column value with token userId

**Describe the solution you'd like**
We can check userTenantId for userId column or we should introduce a provider for checking column value

**Describe alternatives you've considered**
currently using a new controller for handling **/notification-users** get api

samarpan-b:
@sandeep-sourcefuse I dont like the solution of adding a new provider for this. Any other solution ?

harshitmitSF:
@sandeep-sourcefuse can you please elaborate the issue? Do you want the userTeanantId in the response body?

sandeep-sourcefuse:
@harshitmitSF there are two issue which i am facing in  `@sourceloop/notification-service` but this is about first

We are using create-notification
```json
{
      "template": {
        "method": "POST",
        "url": "/notifications",
        "headers": {
          "content-type": "application/json",
          "Authorization": "{token}"
        },
        "body": "{body}"
      },
      "functions": {
        "createNotification": ["body", "token"]
      }
},
```

and the body passed as

```ts
this.notificationService.createNotification(bearerToken, {
      ----
      receiver: {
        to: [{id: userTenantId}],
      },
      type: NotificationType.Push,
      ----
});
```

as you can see here we are passing `userTenantId` to send socket notification

but if we request for all notifications list for particular user then the condition check will incorrect as it is defined in

`@sourceloop/notification-service/src/controllers/notification-user.controller.ts`

```ts
 private _createFilterBuilder(
    currentUser: IAuthUserWithPermissions,
    filter: Filter<NotificationUser> = {},
  ) {
    const filterBuilder = new FilterBuilder(filter);
    if (filter) {
      const whereBuilder = new WhereBuilder(filter.where);
      whereBuilder.and([
        {
          userId: currentUser.id,
        },
      ]);
      filterBuilder.where(whereBuilder.build());
    }
    return filterBuilder;
  }
```
In above check we are `userId: currentUser.id,` not `userId: currentUser.userTenantId,


so can we get something to define our condition

harshitmitSF:
[...]

PR #123, opened 2021-03-10
fix(notification-service): added binding key and provider for notification user where builder

akshatdubeysf:
## Description

Added binding key and provider for where clause builder in the notification user get endpoints.

Fixes # (issue)
#122 

## Type of change

## Checklist:

### DIFF
--- a/services/notification-service/src/controllers/notification-user.controller.ts
+++ b/services/notification-service/src/controllers/notification-user.controller.ts
@@ -4,7 +4,6 @@ import {
   Filter,
   repository,
   Where,
-  WhereBuilder,
   FilterBuilder,
 } from '@loopback/repository';
 import {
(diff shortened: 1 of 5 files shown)


services/notification-service/src/controllers/notification-user.controller.ts
```diff
@@ -350,13 +353,7 @@ export class NotificationUserController {
...
     where?: Where<NotificationUser>,
   ) {
-    const whereBuilder = new WhereBuilder(where);
-    whereBuilder.and([
-      {
-        userId: currentUser.id,
-      },
-    ]);
-    return whereBuilder;
+    return this.notificationWhereBuilder(currentUser, where);
```
samarpan-b, line 356: I dont like this idea of exposing this as a provider. Whats the ask here ?

samarpan-b (changes requested):

services/notification-service/src/controllers/notification-user.controller.ts
akshatdubeysf: The issue is that the filter always adds a check on the userId field, so the suggestion in the issue was to provide a way to override this filter in some way.

PR #176, opened 2021-04-26
fix(authentication-service): added a post route for keycloak login due to a possible security issue

akshatdubeysf:
## Description

Possible security issue if secret is passed via query params, hence added a post route for keycloak
login

## Type of change

## Checklist:


services/authentication-service/src/modules/auth/login.controller.ts
```diff
@@ -446,6 +455,45 @@ export class LoginController {
...
+      callbackURL: process.env.KEYCLOAK_CALLBACK_URL, //'http://localhost:3001/auth/keycloak-auth-redirect',
+      authorizationURL: `${process.env.KEYCLOAK_HOST}/auth/realms/${process.env.KEYCLOAK_REALM}/protocol/openid-connect/auth`,
+      tokenURL: `${process.env.KEYCLOAK_HOST}/auth/realms/${process.env.KEYCLOAK_REALM}/protocol/openid-connect/token`,
+      userInfoURL: `${process.env.KEYCLOAK_HOST}/auth/realms/${process.env.KEYCLOAK_REALM}/protocol/openid-connect/userinfo`,
+    },
+    keycloakBodyGen,
+  )
+  @authorize({permissions: ['*']})
+  @post('/auth/keycloak', {
+    responses: {
```
samarpan-b, line 475: add description to this API

samarpan-b (changes requested):

services/authentication-service/src/modules/auth/login.controller.ts
akshatdubeysf: Done

services/authentication-service/src/modules/auth/login.controller.ts
```diff
@@ -474,6 +524,10 @@ export class LoginController {
       },
     },
   })
+
```
samarpan-b, line 527: remove space and use deprecated for open api schema

services/authentication-service/src/modules/auth/login.controller.ts
```diff
@@ -446,6 +455,46 @@ export class LoginController {
...
+        description: 'Keycloak Token Response',
+        content: {
+          [CONTENT_TYPE.JSON]: {
+            schema: {'x-ts-type': TokenResponse},
+          },
+        },
+      },
+    },
+  })
+  async postloginViaKeycloak(
```
samarpan-b, line 487: rename postLoginViaKeycloak

samarpan-b (changes requested):

services/authentication-service/src/modules/auth/login.controller.ts
```diff
@@ -79,6 +81,14 @@ const keycloakQueryGen = (req: Request) => {
   };
 };
 
+const keycloakBodyGen = (req: Request) => {
+  return {
+    state: Object.keys(req.body)
+      .map(key => `${key}=${req.body[key]}`)
```
sf-kansara, line 87: Should we only pass `client_id`?

services/authentication-service/src/modules/auth/login.controller.ts
```diff
@@ -446,6 +456,46 @@ export class LoginController {
...
+          },
+        },
+      },
+    },
+  })
+  async postLoginViaKeycloak(
+    @requestBody({
+      content: {
+        'application/x-www-form-urlencoded': {
+          schema: {type: 'object'},
```
sf-kansara, line 492: We can have a minimal `jsonSchema` here.

services/authentication-service/src/modules/auth/login.controller.ts
akshatdubeysf: You mean instead of req we pass only the client_id and return a string for that? Instead of mapping over the whole body? I did this just to be consistent with the other method does the same thing for query params.

services/authentication-service/src/modules/auth/login.controller.ts
akshatdubeysf: Added.

services/authentication-service/src/modules/auth/login.controller.ts
samarpan-b: yes. we only need to honor client id. refer -> https://github.com/sourcefuse/rakuten-pms-api/commit/6cd905231d549bcfd42f705a244bc9befd2944e7

akshatdubeysf: Oh got it, will do it.

akshatdubeysf: Done

services/authentication-service/src/modules/auth/login.controller.ts
```diff
@@ -73,9 +75,13 @@ const queryGen = (req: Request) => {
...
 const keycloakQueryGen = (req: Request) => {
   return {
-    state: Object.keys(req.query)
-      .map(key => `${key}=${req.query[key]}`)
-      .join('&'),
+    state: `client_id=${req.query.client_id}`,
+  };
+};
+
+const keycloakBodyGen = (req: Request) => {
```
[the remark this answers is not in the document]
akshatdubeysf: Oh right.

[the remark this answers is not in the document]
akshatdubeysf: But one is works for query params and the other for body?

samarpan-b (approved):

PR #253, opened 2021-07-12
fix(core): fix facades bearer verifier to decouple from auth service

### LINKED ISSUE #252: Facades bearer verifier is tight coupling all facade microservices with auth service
**Describe the bug**
FacadesBearerTokenVerifyProvider - This is tight coupling auth service to all other facades which are using this verify bearer token. This is an anti pattern.

**Expected behavior**
Facades should be self sufficient enough to verify bearer token without depending on another service. This way they can also be completely independent.

**Additional context**
We need to ensure revoked tokens checking still works here.

samarpan-b:
BREAKING CHANGE: authService url removed from BearerVerifierConfig and now there is a need for a data source with AuthCacheSourceName

## Description

Decouple facades bearer verifier from auth service

Fixes #252 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:


packages/core/src/components/bearer-verifier/providers/facades-bearer-token-verify.provider.ts
```diff
@@ -1,27 +1,48 @@
...
+        throw new HttpErrors.Unauthorized('TokenRevoked');
+      }
+
+      let user: IAuthUserWithPermissions;
       try {
-        if (!req || !req.headers) {
-          throw new HttpErrors.Unauthorized(AuthErrorKeys.TokenInvalid);
-        }
-        return await this.authService.getme(req.headers.authorization);
+        user = verify(token, process.env.JWT_SECRET as string, {
```
akshatdubeysf, line 30: If the facades provider is doing the same verification as ServicesBearerTokenVerifyProvider, then why the need for two different providers?

samarpan-b: No its not same. Facades verifier has a reference to revoked token repository and it requires a keyvalue datasource created like redis. Services one is only doing jwt verify, no other dependency.

akshatdubeysf (approved):

yeshamavani (approved):

PR #426, opened 2021-11-23
feat(core): add join mixin for repositories

### LINKED ISSUE #414: Support of JOINs in repository calls
As of now, including a related model in the repository calls results in two DB hits, there should be a way to avoid this using joins.

There should be an optional parameter that when provided should fetch the related models using JOINS instead of separate DB calls - 

repo.find({include: [{relation:'child', join:true}]});

**Describe alternatives you've considered**
Other options that were considered were contributing to the loopback connector repo, but that would be time consuming, so  could try an extension for repositories instead.

akshatdubeysf:
## Description

Add join mixin for repositories

Fixes #414 

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## Checklist:


yeshamavani:
Taken the sequlize approach.

PR #459, opened 2022-01-12
fix(authentication-service): change response of signup and forget password

### LINKED ISSUE #457: Avoid token in response for signup and forgot password
**Describe the bug**
In the current implementation, the signup and forgot create and provide token in responses and these two APIs are open to users.

**Expected behavior**
The token should be provided by the API in response, and should be used internally only.

**Additional context**
Can use providers to solve this, giving a provider for each of these two actions

akshatdubeysf:
BREAKING CHANGE:
removed response body and changed response code of signup-token and forget-password apis

## Description

change response of signup and forget password, add validation for internal user during forget and reset password

Fixes #457 

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [X] New tests added or existing tests modified to cover all changes


packages/core/src/enums/auth-provider.enum.ts
```diff
@@ -0,0 +1,4 @@
+export enum AuthProvider {
+  INTERNAL = 'internal',
+  KEYCLOAK = 'keycloak',
```
samarpanB, line 3: add other providers too

services/authentication-service/.env.defaults
```diff
@@ -27,4 +27,22 @@ REDIS_URL=
 REDIS_PASSWORD=a
 REDIS_DATABASE=a
 FORGOT_PASSWORD_LINK_EXPIRY=30
-REQUEST_SIGNUP_LINK_EXPIRY=30
\ No newline at end of file
+REQUEST_SIGNUP_LINK_EXPIRY=30
+DB_HOST=
```
samarpanB, line 31: remove these

services/authentication-service/src/controllers/signup-request.controller.ts
```diff
@@ -66,11 +66,7 @@ export class SignupRequestController {
       algorithm: 'HS256',
     });
 
-    return new SignupRequestResponseDto({
-      code: token,
-      expiry: expiryDuration,
-      email: signUpRequest.email,
-    });
+    await handler(token, expiryDuration, signUpRequest.email);
```
samarpanB, line 69: send it as DTO only, so that it can be managed easily in future

services/authentication-service/src/controllers/forget-password.controller.ts
```diff
@@ -81,38 +70,27 @@ export class ForgetPasswordController {
...
+      process.env.FORGOT_PASSWORD_LINK_EXPIRY ?? '1800',
+    );
+    const token = jwt.sign(codePayload, process.env.JWT_SECRET as string, {
+      expiresIn: expiryDuration,
+      audience: req.client_id,
+      subject: user.username.toLowerCase(),
+      issuer: process.env.JWT_ISSUER,
+      algorithm: 'HS256',
+    });
+    await forgetPasswordHandler(token, expiryDuration, user.email, user);
```
samarpanB, line 93: use the same DTO object

samarpanB (changes requested):

services/authentication-service/.env.defaults
akshatdubeysf: But the build fails as they are added in the env schema

packages/core/src/enums/auth-provider.enum.ts
akshatdubeysf: done

services/authentication-service/src/controllers/forget-password.controller.ts
```diff
@@ -81,38 +71,32 @@ export class ForgetPasswordController {
...
+      audience: req.client_id,
+      subject: user.username.toLowerCase(),
+      issuer: process.env.JWT_ISSUER,
+      algorithm: 'HS256',
+    });
+    await forgetPasswordHandler({
+      code: token,
+      expiry: expiryDuration,
+      email: user.email,
+      user: user,
```
samarpan-b, line 98: why do we need user model here ? we should not send that.

samarpan-b (changes requested):

sf-kansara (approved):

services/authentication-service/src/controllers/forget-password.controller.ts
akshatdubeysf: It was part of the ForgetPasswordDto, can be removed if not required.

akshatdubeysf: done

samarpan-b (approved):

Issue #528, opened 2022-02-16
@sourceloop/cli should provide a command to update existing microservice


akshatdubeysf:
**Is your feature request related to a problem? Please describe.**
The CLI should provide a command to update already existing microservices.

**Describe the solution you'd like**
We can provide a command like `sl update microservice <name>` that will perform the steps required to update a microservic

PR #835, opened 2022-06-14
feat(cli): add new update command to cli

yeshamavani:
gh-528

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/packages/cli/src/generators/update/index.ts
+++ b/packages/cli/src/generators/update/index.ts
@@ -1,9 +1,4 @@
-import {
-  AnyObject,
-  DependencyType,
-  PackageDependencies,
-  UpdateOptions,
-} from '../../types';
+import {AnyObject, PackageDependencies, UpdateOptions} from '../../types';
 import BaseUpdateGenerator from '../../update-generator';
 import {join} from 'path';
 import {readdirSync} from 'fs';
--- a/packages/cli/src/generators/update/index.ts
+++ b/packages/cli/src/generators/update/index.ts
@@ -1,4 +1,9 @@
-import {AnyObject, DependencyType, UpdateOptions} from '../../types';
+import {
+  AnyObject,
+  DependencyType,
+  PackageDependencies,
+  UpdateOptions,
+} from '../../types';
 import BaseUpdateGenerator from '../../update-generator';
 import {join} from 'path';
 import {readdirSync} from 'fs';
@@ -85,11 +90,10 @@ export default class UpdateGenerator extends BaseUpdateGenerator<UpdateOptions> 
       ];
       const answer = await this.prompt(prompts);
       if (answer && answer.decision === 'continue') {
-        return false;
+        return;
       }
       if (answer && answer.decision === 'upgrade') {
         await this._updateDependencies();
-        return true;
       }
     }
   }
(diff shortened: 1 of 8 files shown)


packages/cli/src/generators/update/index.ts
```diff
@@ -0,0 +1,281 @@
...
+      return false;
+    }
+  }
+
+  private async _checkDependencies() {
+    const packageJson = this.fs.readJSON(
+      this.destinationPath(packageJsonFile),
+    ) as AnyObject;
+
+    // const pkgDeps: {
```
akshatdubeysf, line 110: ???

packages/cli/src/generators/update/index.ts
```diff
@@ -0,0 +1,281 @@
...
+      }
+      if (answer && answer.decision === 'upgrade') {
+        await this._updateDependencies();
+        return true;
+      }
+    }
+  }
+
+  private _compare(tempDependency: string, dependency: string) {
+    if (dependency !== tempDependency) {
```
akshatdubeysf, line 98: return dependency !== tempDependency

akshatdubeysf (changes requested):

packages/cli/src/generators/update/index.ts
```diff
@@ -0,0 +1,252 @@
...
+
+const packageJsonFile = 'package.json';
+
+export default class UpdateGenerator extends BaseUpdateGenerator<UpdateOptions> {
+  constructor(public args: string[], public opts: UpdateOptions) {
+    super(args, opts);
+  }
+
+  initializing() {
+    if (!this.fs.exists(join(this.destinationPath(), 'lerna.json'))) {
```
samarpan-b, line 19: why is this needed at rot only ?

packages/cli/src/generators/update/index.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    for (const type of types) {
+      this.destinationRoot(monoRepo);
+      const folders = this._getDirectories(
+        this.destinationRoot(join('.', type)),
+      );
+      for (const folder of folders) {
+        this.destinationRoot(join(monoRepo, type, folder));
+
+        await super.checkLoopBackProject();
```
samarpan-b, line 36: await inside loop is simply bad programming

packages/cli/src/generators/update/index.ts
```diff
@@ -0,0 +1,252 @@
...
+  async updateAllProjects() {
+    const types = ['facades', 'services', 'packages'];
+    const monoRepo = this.destinationPath();
+
+    for (const type of types) {
+      this.destinationRoot(monoRepo);
+      const folders = this._getDirectories(
+        this.destinationRoot(join('.', type)),
+      );
+      for (const folder of folders) {
```
samarpan-b, line 33: cant we reduce time complexity here ? Its O(n^2) right now. Try reducing it to O(n)

packages/cli/src/generators/update/index.ts
```diff
@@ -0,0 +1,252 @@
...
+      }
+    }
+
+    // lerna bootstrap at the end
+    this.destinationRoot(monoRepo);
+    this.spawnCommandSync('lerna', ['bootstrap']);
+  }
+
+  private _getDirectories(folderPath: string) {
+    return readdirSync(folderPath, {withFileTypes: true})
```
samarpan-b, line 57: shouldn't we use async method here ?

packages/cli/src/generators/update/index.ts
```diff
@@ -0,0 +1,252 @@
...
+      const devDep = pkgDeps.devDependencies[d];
+      const peerDep = pkgDeps.peerDependencies[d];
+      if (!dep && !devDep && !peerDep) continue;
+
+      const tempDependency = tempDeps[d];
+      if (dep && this._compare(tempDependency, dep)) {
+        depsToUpdate.dependencies[d] = tempDependency;
+        found = true;
+      }
+      if (devDep && this._compare(tempDependency, devDep)) {
```
samarpan-b, line 120: this._compare is invoked multiple times in the same fn. We can invoke once and use its value. This improves performance of the code.

packages/cli/src/types.ts
```diff
@@ -7,6 +7,18 @@ export interface ProjectInfo {
     [key: string]: string;
   };
 }
+
+// sonarignore:start
+export interface DependencyType {
```
samarpan-b, line 12: move types specific to sl update  command into separate file.

samarpan-b (changes requested):

packages/cli/src/generators/update/index.ts
yeshamavani: we will need this as not able to achieve desired flow without it

packages/cli/src/generators/update/index.ts
yeshamavani: actually the outer loop is fixed so it is like m*n = 3n,

packages/cli/src/generators/update/index.ts
yeshamavani: removed the function checking directly

packages/cli/src/types.ts
yeshamavani: done

akshatdubeysf (approved):

PR #916, opened 2022-07-18
fix(cache): override findOne method and add caching to it

barleendhaliwal:
## Description

- Caching was not done for findOne method. Fixed that. 
- Find One method can return null as response. Null responses are also stored in cache. Changed types for strategy interface accordingly.

Fixes # (gh-0)

## Type of change

## Checklist:

### DIFF
--- a/packages/cache/src/mixins/cache.mixin.ts
+++ b/packages/cache/src/mixins/cache.mixin.ts
@@ -9,6 +9,7 @@ import {
   Filter,
   FilterExcludingWhere,
   JugglerDataSource,
+  Options,
 } from '@loopback/repository';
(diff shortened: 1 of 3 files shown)


packages/cache/src/mixins/cache.mixin.ts
```diff
@@ -120,6 +121,38 @@ export class CacheManager {
         return finalResult;
       }
 
+      /* eslint-disable-next-line @typescript-eslint/ban-ts-comment */
+      // @ts-ignore
+      async findOne(
```
samarpan-b, line 126: Then why don't we do findbyid as well? id can be the key there.

samarpan-b (changes requested):

PR #1125, opened 2022-12-14
fix(chore): added if condition for createall db call

### LINKED ISSUE #1124: createAll method needs array length check
**Describe the bug**
Since the latest loopback release 
if an empty array is passed to createAll method it throws an error, so a safety check is needed.
**To Reproduce**
1. Try passing an empty array to the createAll method 
2. Check error

**Expected behavior**
This should be handled at our end.

yeshamavani:
Breaking Change 
if the array is not passed error will be thrown 
gh-1124

## Description

Added the if condition to check the entity length before calling the repo.createAll call
to avoid error.

Fixes #1124

## Type of change

## Checklist:

### DIFF
--- a/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
+++ b/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
@@ -57,6 +57,9 @@ export abstract class DefaultTransactionalUserModifyRepository<
       entity.createdBy = uid ?? '';
       entity.modifiedBy = uid ?? '';
     });
+    if (entities.length <= 0) {
+      throw new HttpErrors.BadRequest('Entity cannot be empty');
+    }
     return super.createAll(entities, options);
   }
 
--- a/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
+++ b/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
@@ -57,7 +57,7 @@ export abstract class DefaultTransactionalUserModifyRepository<
       entity.createdBy = uid ?? '';
       entity.modifiedBy = uid ?? '';
     });
-    if (entities.length <= 0) {
+    if (!entities.length) {
       throw new HttpErrors.BadRequest('Entity cannot be empty');
     }
     return super.createAll(entities, options);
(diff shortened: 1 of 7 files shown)


packages/core/src/repositories/default-user-modify-crud.repository.base.ts
```diff
@@ -58,6 +58,9 @@ export class DefaultUserModifyCrudRepository<
       entity.createdBy = uid ?? '';
       entity.modifiedBy = uid ?? '';
     });
+    if (entities.length <= 0) {
```
akshatdubeysf, line 61: Why not just check entities.length

services/chat-service/src/controllers/attachment-file.controller.ts
```diff
@@ -96,6 +97,9 @@ export class AttachmentFileController {
     attachFileDto.attachmentFiles.forEach(messageFile => {
       attachFiles.push(new AttachmentFile(messageFile));
     });
+    if (attachFiles.length <= 0) {
```
akshatdubeysf, line 100: If is at repo level, why is it required here?

akshatdubeysf (changes requested):

services/chat-service/src/controllers/attachment-file.controller.ts
yeshamavani: will remove this

packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
```diff
@@ -57,6 +57,9 @@ export abstract class DefaultTransactionalUserModifyRepository<
       entity.createdBy = uid ?? '';
       entity.modifiedBy = uid ?? '';
     });
+    if (!entities.length) {
```
akshatdubeysf, line 60: This should be breaking change

akshatdubeysf (changes requested):

yeshamavani:
This will be released when there is a breaking change

yeshamavani:
Raised another PR with same changes to avoid merge conflicts - https://github.com/sourcefuse/loopback4-microservice-catalog/pull/1881

PR #1193, opened 2023-01-11
docs(core): add mkdocs and typedoc configuration

### LINKED ISSUE #1192: Documentation site

shubhamp-sf:
image

Preview: https://loopback4-microservice-catalog.vercel.app/
1. Added and configured mkdocs and typedoc for later mentioned services and packages in order to export API references.
2. Added "Docs" workflow script that rebuilds API References on push to master branch and updated compiled html files on `gh-pages` branch.
3. By default typdoc won't include other markdown files such as `openapi.md` so to use it I've used custom script that tweaks the formatting of that files and copy it into the folder when building docs.
4. It's important to follow common formatting standards in README files, failure to which will build errors while exporting html files, one common mistake is:
     - Not using backticks (\`\`) while typing out decorators/any code term. for example writing @authentication will fail as it considers this as a tsdoc tag. So this needs to be properly enclosed within them.
 
> KNOWN ISSUE: While hosting this on github pages the default links are broken as mkdocs adds up "../" to all links. So It's better to have a dedicated domain for it like it worked with vercel here: https://loopback4-microservice-catalog.vercel.app/ .

---

Currently, API References for following services and packages is being exported:
Services
- Audit service
- Authentication service
- Bpmn service
- Chat service
- Feature toggle service
- In mail service
- Notification service
- Payment service
- Reporting service
- Scheduler service
- Search service
- User tenant service
- Video conferencing service

**Packages**
- Cache
- Core
- Feature toggle

---

Closes #1192


jamescrowley321 (approved):

export-typedocs.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+lerna exec --scope '{@sourceloop/core,@sourceloop/audit-service,@sourceloop/authentication-service,@sourceloop/feature-toggle,@sourceloop/bpmn-service,@sourceloop/chat-service,@sourceloop/feature-toggle-service,@sourceloop/in-mail-service,@sourceloop/notification-service,@sourceloop/ocr-s3-service,@sourceloop/ocr-service,@sourceloop/payment-service,@sourceloop/reporting-service,@sourceloop/scheduler-service,@sourceloop/search-service,@sourceloop/user-tenant-service,@sourceloop/video-conferencing-service,@sourceloop/cache}' -- npx typedoc --logLevel Error --excludeExternals
```
akshatdubeysf, line 3: Why not write a TS/JS script? that would be easier to maintain and understand? We could have logic to read all packages so addition of a service or package doesn't require updating this script (assuming we also have an option to filter out the non-required services/packages within this script)

services/authentication-service/README.md
```diff
@@ -118,69 +124,197 @@ npm i @sourceloop/authentication-service
...
-| `KEYCLOAK_REALM` | N | | Realm (tenant) in Keycloak |
-
-| `KEYCLOAK_CLIENT_ID` | N | | Keycloak client ID for the service |
-
-| `KEYCLOAK_CLIENT_SECRET` | N | | Keycloak client secret for the service |
-
-| `KEYCLOAK_CALLBACK_URL` | N | | Keycloak callback URL for the client configuration in Google |
-
-| `HTTPS_PROXY` | N | | Https proxy url for keycloak auth |
+<table>
```
akshatdubeysf, line 127: This same document is also rendered in npm page for this service, would this table work there?

akshatdubeysf (changes requested):

services/authentication-service/README.md
akshatdubeysf: Then this should be fine 🙂

export-typedocs.sh
shubhamp-sf: Surely it would but I feel `.sh` file is a better place when we just need to run some commands one after another.
The only down side is maintaining bulky scope parameter, I will see if this can be stripped out elegantly.

What do you think about adding additional `command` besides bootstrap etc. in `lerna.json`? and just running `lerna run export-typedocs` here or in root `package.json`?

akshatdubeysf: But these are not simple independent scripts na? If a new service is added, it would have to be manually added to every single script, a manual work that can be avoided. Even if lerna allows writing a command, that would still require you to add the list manually, for each sub command or script. Considering all other automations are in js/ts, why do you feel an `.sh` script would be better?  And even if you want to stick to `sh`, you can write this script in a way to avoid manual rework.

akshatdubeysf: In the script you can write logic to read some custom properties from the service/packages package.json na? You write code to read all services/packages and filter them out based on these package.json properties 🙂

shubhamp-sf: Are ha... 

Boss you're so smart! 😎

Will do this. 👍

shubhamp-sf: Done.

shubhamp-sf: I created [export-typedocs.ts](https://github.com/sourcefuse/loopback4-microservice-catalog/pull/1193/files#diff-de1112299a30d65543ae0366c47b252f204fd9d02fd927154f212265211bc240) if you want to review.

An execution overview of `export-typedocs.ts` would be this:
1. It runs the command `lerna ls` to get a list of all public packages.
2. It iterates through each package in the retrieved packages.
3. For each package, it attempts to import the package's package.json file where it reads the optional field `typedoc`.
For API references to work, packages have similar to below object in their package.json file
```json
{
  "typedoc": {
    "config": {
      "entryPoints": [ "src/index.ts" ],
      "out": "services/authentication-service"
    },
    "settings": {
      "includeOpenAPIDoc": true
    }
  }
}
```
***This also eliminated the need of extra `typedoc.json` in every target package 🙂, thanks for the suggestion.***

Can we mark this as resolve now?

jamescrowley321: @shubhamp-sf Can you add that command to the SL CLI and make sure the steps are documented? We can use this same technique elsewhere. Feel free to raise another PR for that.

shubhamp-sf: Sure.

PR #1251, opened 2023-01-26
fix(cache): change clearCache function implementation and add some types

barleendhaliwal:
## Description

Approach for deletion:
- store insertion time for each key along with payload
- on call of clear cache function store the epoch time in a key in redis. key: ${this.prefix}_DELETION_TIME
- when user requests for something which is present in redis - check the deletion time from this key. if insertion time of requested key <= deletion time delete that key from redis.

Fixes gh-0

## Type of change

## Checklist:


packages/cache/src/strategies/redis/redis-cache-strategy.ts
```diff
@@ -63,41 +75,20 @@ export class RedisCacheStrategy<M> implements ICacheStrategy<M> {
...
-    if (!this.getCacheDataSource) {
-      throw new Error(`Please provide value for getCacheDataSource`);
-    }
-  }
-
   async executeRedisCommand(
     command: string,
     args: (string | number)[],
-  ): Promise<Buffer | number | undefined> {
+  ): Promise<RedisConnectorExecuteReturnType> {
```
akshatdubeysf, line 91: Why not take a generic type instead of using as in the parent if the return type is decided by us?

packages/cache/src/strategies/redis/redis-cache-strategy.ts
```diff
@@ -7,54 +7,66 @@ import {HttpErrors} from '@loopback/rest';
...
     try {
       const res = (await this.executeRedisCommand('GET', [key])) as Buffer;
       if (res) {
         result = JSON.parse(decoder.decode(res));
+        const deletionTime = (await this.executeRedisCommand('GET', [
+          this.deletionKey,
+        ])) as Buffer;
+        if (
+          deletionTime &&
+          (result as CacheEntity<M>).insertionTime <=
```
akshatdubeysf, line 44: This would fetch the entire cached data, right? why not do this in a script as we discussed? The script would take a key and can return data only if the key is still valid

akshatdubeysf (changes requested):

packages/cache/src/strategies/redis/redis-cache-strategy.ts
barleendhaliwal: done

packages/cache/src/strategies/redis/redis-cache-strategy.ts
akshatdubeysf: I will do this 🙁

PR #1518, opened 2023-06-27
feat(authentication-service): add filter to include or exclude active users

### LINKED ISSUE #1517: Feature: Add filtering options for active users
Right now the active users API gives the data for a particular date range daily/monthly
but there is no way to filter the data where we can include or exclude only specific users in the active users range.

This is a need in a scenario where in production we have actual users as well as dummy users for testing. 
So while getting the active users list there has to be a way where we can exclude the dummy. 

Provide a flag to mark whether we would like to include or exclude the user. Also a array where we can pass the list of users whom we want to include/exclude.

yeshamavani:
gh-1517

## Description

Provide a flag to mark whether we would like to include or exclude the user. Also a array where we can pass the list of users whom we want to include/exclude.

Fixes #1517 

## Type of change

## Checklist:


services/authentication-service/src/controllers/login-activity.controller.ts
```diff
@@ -179,4 +199,45 @@ export class LoginActivityController {
...
+
+  private async buildActiveUsersFilter(
+    filter: ActiveUsersFilter,
+  ): Promise<AnyObject> {
+    let actorIds: string[] = [];
+    if (filter.userIdentity === UserIdentity.ACTOR_ID) {
+      actorIds = filter.userIdentifier;
+    } else {
+      // check if actor is userTenantId
+      if (this.actorKey === 'id') {
```
samarpan-b, line 211: what is this.actorKey ?

services/authentication-service/src/controllers/login-activity.controller.ts
```diff
@@ -179,4 +199,45 @@ export class LoginActivityController {
 
     return groupByDate;
   }
+
+  private async buildActiveUsersFilter(
+    filter: ActiveUsersFilter,
+  ): Promise<AnyObject> {
+    let actorIds: string[] = [];
```
samarpan-b, line 206: Can we make this function as a provider so that consumers can configure it also, if they want ?

samarpan-b (changes requested):

services/authentication-service/src/controllers/login-activity.controller.ts
yeshamavani: since we wanted the actor column to be configurable user can set the actionKey to be id== userId
or userTenantId or any other field in AuthUser

services/authentication-service/src/controllers/login-activity.controller.ts
yeshamavani: sure

services/authentication-service/src/controllers/login-activity.controller.ts
samarpan-b: but why is it a class variable ?

yeshamavani:
Raised another PR with same changes as this one is very old and uses lerna and not npm workspaces.
https://github.com/sourcefuse/loopback4-microservice-catalog/pull/2049

PR #1487, opened 2023-06-21
feat(task-service): new task management microservice 

antriksha102:
Task Service 

## Description

- [x] The requirement is to build a reusable, customisable and workflow based task service which creates some actionable tasks based upon various events happening via different micro-services in a distributed system.

Fixes #1484 

## Type of change
- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] Performed a self-review of my own code
- [x] npm test passes on your machine
- [x] Code conforms with the style guide

PR #1681, opened 2023-09-04
fix(task-service-example): fix migrations and add camunda bpmn provider binding

shubhamp-sf:
## Description

MIGRATION CHANGE:
migration-20230809100541- changed the duplicate primary key constraint


shubhamp-sf:
Closing as the changes are already incorporated in https://github.com/sourcefuse/loopback4-microservice-catalog/pull/1487

Issue #1972, opened 2024-02-12
task-service: add way to delete process instances on Camunda


Swati0077:
**Is your feature request related to a problem? Please describe.**
Whenever we are deleting  tasks/user-tasks they are deleted from Database but For example if we are using Camunda the process instances are in running state on Camunda Server. So they should be deleted as well.

**Describe the solution you'd like**
1. Add delete API for process instance deletion in Camunda.
2. When deleting the tasks, also delete the process instances.

**Describe alternatives you've considered**
1. With the help of Camunda process instance deletion API.

PR #2028, opened 2024-03-11
feat(task-service): delete process instances on Camunda for deleted tasks

Swati0077:
## Description

1. Delete process instances on Camunda for deleted tasks.
2. Remove ? to mark externalId as mandatory field.
3. Add Camunda's process instances delete API to delete for process instances from Camunda 

Fixes GH-1972

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist:

- [x] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/services/task-service/src/controllers/task.controller.ts
+++ b/services/task-service/src/controllers/task.controller.ts
@@ -1,3 +1,4 @@
+import {inject, service} from '@loopback/core';
 import {
   Count,
   CountSchema,
@@ -16,6 +17,8 @@ import {
 } from '@loopback/rest';
 import {
   CONTENT_TYPE,
+  ILogger,
+  LOGGER,
   OPERATION_SECURITY_SPEC,
   STATUS_CODE,
 } from '@sourceloop/core';
@@ -28,6 +31,7 @@ import {
   TaskWorkFlowRepository,
   UserTaskRepository,
 } from '../repositories';
+import {CamundaService} from '../services';
 
 const baseUrl = 'tasks';
 
@@ -39,6 +43,9 @@ export class TaskController {
     private readonly userTaskRepository: UserTaskRepository,
     @repository(TaskWorkFlowRepository)
     private readonly taskWorkflowMapping: TaskWorkFlowRepository,
+    @service(CamundaService)
+    private readonly camundaService: CamundaService,
+    @inject(LOGGER.LOGGER_INJECT) private readonly logger: ILogger,
   ) {}
 
   @authenticate(STRATEGY.BEARER)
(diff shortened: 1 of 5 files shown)


yeshamavani:
Will be a breaking change

services/task-service/src/models/task.model.ts
```diff
@@ -102,5 +102,5 @@ export class Task<TS = TaskStatus> extends UserModifiableEntity<Task> {
     name: 'external_id',
     type: 'string',
   })
-  externalId?: string;
+  externalId: string;
```
samarpan-b, line 105: Why are we changing this ? This will require DB migration as well, which I would like to avoid.

services/task-service/src/controllers/task.controller.ts
```diff
@@ -175,16 +182,27 @@ export class TaskController {
...
+
+      const processInstanceIds = tasks.map(task => task.externalId);
+
       if (tasks.length) {
         await this.userTaskRepository.deleteAllHard({
           taskId: {
             inq: tasks.map(task => task.id!),
           },
         });
+        this.camundaService
```
samarpan-b, line 196: This needs to be done in deletebyid API as well. Also check if this should be done on task completion API or not.

samarpan-b (changes requested):

services/task-service/src/models/task.model.ts
Swati0077: Yes, considering the wide variety of use cases and discussion with @akshatdubeysf  we can't limit this to be a mandatory field.
So I have changed to the existing code , making it non mandatory

services/task-service/src/controllers/task.controller.ts
Swati0077: Added for deleteByID as well..
For completion API , its already working fine.

services/task-service/src/services/camunda.service.ts
```diff
@@ -37,6 +37,20 @@ export class CamundaService {
     );
   }
 
+  async deleteProcessInstances(ids: (string | undefined)[]) {
+    if (ids) {
+      return Promise.all(
+        ids.map(id =>
+          this.http.delete(`${this.baseUrl}/process-instance/${id}`, {
+            query: {
```
samarpan-b, line 45: skipCustomListeners should also be set to true

samarpan-b: https://docs.camunda.org/manual/7.13/reference/rest/process-definition/delete-process-definition/#query-parameters

Swati0077: added

samarpan-b (approved):

PR #2159, opened 2024-08-30
fix(core): createdon modifiedon can not be changed via input

### LINKED ISSUE #2158: @sourceloop/core: CreatedOn and ModifiedOn columns should not be data driven
**Describe the bug**
CreatedOn and ModifiedOn columns should not be data driven
These values should always be a the current date and user should never be able to change its value based on input.

**To Reproduce**
1. Hit any post API where the base model entity is UserModifiableEntity
2. Pass the createdOn and modifiedOn date values as some backdated data 
3. Observe the entry in database
4. Instead of current date the values will be one provided in input

**Expected behavior**
Should always be the current date time.

yeshamavani:
gh-2158

## Description

createdon modifiedon can not be changed via input
even if its passed in the input we will delete the fields and update it from db

Fixes #2158 

## Type of change

## Checklist:

### DIFF
--- a/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
+++ b/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
@@ -72,7 +72,7 @@ export abstract class DefaultTransactionalUserModifyRepository<
     const uid = currentUser?.userTenantId ?? currentUser?.id;
     entity.modifiedBy = uid;
     delete entity.createdOn;
-    delete entity.modifiedOn;
+    entity.modifiedOn = new Date();
     return super.save(entity, options);
   }
 
--- a/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
+++ b/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
@@ -41,8 +41,8 @@ export abstract class DefaultTransactionalUserModifyRepository<
       throw new HttpErrors.Forbidden(AuthErrorKeys.InvalidCredentials);
     }
     const uid = currentUser?.userTenantId ?? currentUser?.id;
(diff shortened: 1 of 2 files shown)


packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
```diff
@@ -41,6 +41,8 @@ export abstract class DefaultTransactionalUserModifyRepository<
       throw new HttpErrors.Forbidden(AuthErrorKeys.InvalidCredentials);
     }
     const uid = currentUser?.userTenantId ?? currentUser?.id;
+    delete entity.createdOn;
```
samarpan-b, line 44: use new Date()

samarpan-b (changes requested):

packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
yeshamavani: for create() methods me can do new Date() 
but for update() type of method we will have to delete the createdOn entry ?
will that work

yeshamavani: done

packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
```diff
@@ -67,6 +71,8 @@ export abstract class DefaultTransactionalUserModifyRepository<
     }
     const uid = currentUser?.userTenantId ?? currentUser?.id;
     entity.modifiedBy = uid;
+    delete entity.createdOn;
+    delete entity.modifiedOn;
```
samarpan-b, line 75: use new Date()

PR #2188, opened 2024-10-25
feat(core): restrict modification of createdon and modifiedon

### LINKED ISSUE #2158: @sourceloop/core: CreatedOn and ModifiedOn columns should not be data driven

yeshamavani:
gh-2158

## Description

restrict modification of createdon and modifiedon
with this the existing functionality will be same 
all the user needs to do is pass on the property to restrict otherwise it will allow you to pass createdOn and modifiedOn via API as well(current behaviour)

Fixes #2158

## Type of change

## How Has This Been Tested?

Added it to an existing repository and tested it 

## Checklist:

### DIFF
--- a/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
+++ b/packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
@@ -16,6 +16,7 @@ import {AuthErrorKeys} from 'loopback4-authentication';
 import {DefaultTransactionSoftCrudRepository} from 'loopback4-soft-delete';
 import {IAuthUserWithPermissions} from '../components';
 import {UserModifiableEntity} from '../models';
+import {RepositoryOverridingOptions} from '../types';
 
 export abstract class DefaultTransactionalUserModifyRepository<
   T extends UserModifiableEntity,
@@ -34,6 +35,8 @@ export abstract class DefaultTransactionalUserModifyRepository<
     super(entityClass, dataSource);
   }
 
+  public overridingOptions?: RepositoryOverridingOptions;
+
   async create(entity: DataObject<T>, options?: Options): Promise<T> {
     let currentUser = await this.getCurrentUser();
     currentUser = currentUser ?? options?.currentUser;
@@ -43,6 +46,10 @@ export abstract class DefaultTransactionalUserModifyRepository<
     const uid = currentUser?.userTenantId ?? currentUser?.id;
     entity.createdBy = uid;
     entity.modifiedBy = uid;
+    if (this.overridingOptions?.restrictDateModification) {
+      delete entity.createdOn;
+      delete entity.modifiedOn;
+    }
     return super.create(entity, options);
   }
 
@@ -56,6 +63,10 @@ export abstract class DefaultTransactionalUserModifyRepository<
     entities.forEach(entity => {
       entity.createdBy = uid ?? '';
       entity.modifiedBy = uid ?? '';
+      if (this.overridingOptions?.restrictDateModification) {
+        delete entity.createdOn;
+        delete entity.modifiedOn;
+      }
     });
     return super.createAll(entities, options);
   }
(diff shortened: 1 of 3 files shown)


a-ganguly (approved):
restrict modification of createdon and modifiedon

packages/core/README.md
```diff
@@ -570,6 +570,29 @@ export class UserRepository extends DefaultUserModifyCrudRepository<
   )
 ```
 
+The repository by default allows to pass back dated or future dates for createdOn and modifiedOn values but to restrict the manual date modification we can pass overridingOptions like this
```
samarpan-b, line 573:
```suggestion
The repository by default does not restrict setting up of createdOn and modifiedOn through API or external sources. However, you can restrict it using the overridingOptions like this
```

packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
```diff
@@ -43,6 +46,10 @@ export abstract class DefaultTransactionalUserModifyRepository<
     const uid = currentUser?.userTenantId ?? currentUser?.id;
     entity.createdBy = uid;
     entity.modifiedBy = uid;
+    if (this.overridingOptions?.restrictDateModification) {
```
samarpan-b, line 49: why cant we use the second parameter of the method options?: Options. That ways its consistent and easy to implement.

packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
```diff
@@ -34,6 +35,8 @@ export abstract class DefaultTransactionalUserModifyRepository<
     super(entityClass, dataSource);
   }
 
+  public overridingOptions?: RepositoryOverridingOptions;
```
samarpan-b, line 38: I dont like this approach. This will mean they will have to do it for all the repo files. Why ? We need to support both mechanism. One using options and another using a global config.

samarpan-b (changes requested):

packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
yeshamavani: Using the method's option parameter user will have to make changes in all the method calls 
but with this only one change at repo level

samarpan-b: But this introduces a new param whereas we already have such options in method param. The second param is for that purpose only. Moreover, using the method options param gives more flexibility at runtime.

packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
yeshamavani: Tried passing an optional config to this method but it still requires it from the derived class no mater what 
making it a Breaking change - hence used this approach

packages/core/README.md
yeshamavani: done

packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
yeshamavani: Now using the coreConfig for decision making

packages/core/src/repositories/default-transactional-user-modify-repository.base.ts
yeshamavani: Now using the coreConfig for decision making

samarpan-b:
This was already resolved in this - https://github.com/sourcefuse/loopback4-microservice-catalog/pull/2159/changes

yeshamavani:
> This was already resolved in this - https://github.com/sourcefuse/loopback4-microservice-catalog/pull/2159/changes

we reverted that and discussed this approach 

reverted PR - https://github.com/sourcefuse/loopback4-microservice-catalog/commit/1307aa05de60b27ef8be129ed1a662fb3c1f53b6

samarpan-b:
I dont think this approach is right. Why did we revert the earlier PR ? Remind me that please

yeshamavani:
> I dont think this approach is right. Why did we revert the earlier PR ? Remind me that please

You said it break existing functionality and that we should provide backward compatibility, this was VAPT issue.

PR #2247, opened 2025-03-13
fix(file-utils): remove direct export of clamav to avoid direct dep

akshatdubeysf:
## Description

- remove direct export of clamav to avoid direct dependency on clamav to avoid making it a required dependency
- it can now be used only through the sub-package
- better handling of the multipart request schema for openapi-spec
- improved stream handling slightly, but there is still scope for improvement 
- remove file on error in any of the validators

## Type of change

## Checklist:

### DIFF
--- a/packages/file-utils/src/services/validators/index.ts
+++ b/packages/file-utils/src/services/validators/index.ts
@@ -1,3 +1,2 @@
-export * from '../../sub-packages/clamav/validators/clamav-validator.provider';
 export * from './file-name-validator.provider';
 export * from './file-type-validator.provider';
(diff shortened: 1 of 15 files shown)


packages/file-utils/src/services/validators/index.ts
```diff
@@ -1,3 +1,2 @@
-export * from '../../sub-packages/clamav/validators/clamav-validator.provider';
```
samarpan-b, line 1: If this is removed completely then how will consumer reference clamav-validator ?

akshatdubeysf: It was exported through the clamav sub-package already

samarpan-b (approved):

Issue #2249, opened 2025-03-26
Issue with @sourceloop/cache caching functionality


Akanksha13-dev:
**Describe the bug**
When fetching data with a specific filter, then updating the data and retrieving it again with a different filter, the updated data is fetched correctly. However, when fetching data again with the original filter, it still returns the old (pre-update) data instead of the latest updates.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

PR #2250, opened 2025-03-26
fix(cache): remove uninvalidate as it can undo invalidation in case of subsequent get

akshatdubeysf:
## Description

- removed the `uninvalidate` call as it can undo invalidation on any subsequent get, even with a different filter.
- added test case to cover this exact scenario
- refactored the tests a bit for better timeout handling

Fixes GH-2249

## Type of change

## Checklist:

- [X] New tests added or existing tests modified to cover all changes

### DIFF
--- a/packages/cache/src/__tests__/helpers/test-builder.ts
+++ b/packages/cache/src/__tests__/helpers/test-builder.ts
@@ -13,7 +13,6 @@ type TestCase = {
     secondRepo?: DefaultCrudRepository<Test, number, {}>,
     tick?: TickPromise,
   ) => Promise<void>;
-  timeout?: 3000;
 };
 
 type TestSuite = {
@@ -170,7 +169,6 @@ export function repoTestBuilder(
             // one call for value and one for insertion time
             expect(cacheStoreSetSpy.callCount).to.equal(4);
           },
-          timeout: 3000,
         },
       ],
     },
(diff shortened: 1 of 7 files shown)


packages/cache/src/__tests__/helpers/test-builder.ts
```diff
@@ -673,6 +670,28 @@ export function repoTestBuilder(
...
+          ) {
+            await repo.updateById(1, {name: 'test-new-1'});
+            await flushPromises();
+            await repo.updateById(2, {name: 'test-new-2'});
+            await flushPromises();
+            const updatedLast = await repo.findById(2);
+            await flushPromises();
+            const updatedFirst = await repo.findById(1);
+            await flushPromises();
+            expect(updatedLast).to.match({id: 2, name: 'test-new-2'});
```
samarpan-b, line 691: How is this ensuring this data is from cache ? And how is this catering to what you are mentioning as test title ?

akshatdubeysf: The idea is ->

- first get calls create a cache
- update the records
- find the records again
- with the buggy code, the find for 1 id would return the cached record(not updated) as the updateById for 2 would undo all the invalidations done before it for the same tag

samarpan-b (approved):

a-ganguly (approved):

PR #2369, opened 2025-11-24
fix(cli): expose commands through MCP

piyushsinghgaur1:
fix(cli): expose commands through MCP


rohit-sourcefuse:
## ✅ Re-validation Review - Updated PR Status

### 🎉 **SonarCloud Quality Gate: PASSED** ✅

Great progress! The SonarCloud quality gate is now **passing** with:
- 0 New issues
- 0 Security Hotspots
- 0 Accepted issues

However, there are **critical issues** that still need attention:

---

## 🚨 **Critical Issues Still Present**

### **Issue #1: Unsafe Object Stringification - High Priority** ❌

**Location:** `packages/cli/src/commands/mcp.ts` lines 150-152, 167-169

**Problem:** The code now uses `JSON.stringify()` directly without error handling for circular references or complex objects:

```typescript
message: args
  .map(a => (typeof a === 'string' ? a : JSON.stringify(a)))
  .join(' '),
```

**Why This Breaks:**
- `JSON.stringify()` will **throw an error** when encountering circular references
- No fallback mechanism = MCP client logging will crash
- Copilot already flagged this exact issue in the Files Changed tab

**Fix Required:**
```typescript
message: args
  .map(a => {
    if (typeof a === 'string') return a;
    try {
      return JSON.stringify(a);
    } catch (error) {
      // Handle circular references and other stringification errors
      return '[Object with circular reference]';
    }
  })
  .join(' '),
```

**Impact:** Without this fix, logging any object with circular references will cause the MCP server to crash.

---

### **Issue #2: Incorrect Log Level - Medium Priority** ⚠️

**Location:** `packages/cli/src/commands/mcp.ts` line 149

**Problem:** Error logging level changed from 'error' to 'debug':

```typescript
// Changed from:
level: 'error',  // ❌ Removed
// To:
level: 'debug',  // ✅ Current (WRONG)
```

**Why This Is Wrong:**
- Real errors will be logged as debug messages
- Makes it extremely hard to identify actual issues in production
- Copilot explicitly flagged this issue with a detailed explanation

**Fix Required:**
```typescript
level: 'error',  // Restore original error level
```

---

### **Issue #3: Optional Argument Validation Broken - Medium Priority** ⚠️

**Location:** `packages/cli/src/commands/mcp.ts` line 182

**Problem:** The `argToZod` function now always returns `option` instead of handling required vs optional:

```typescript
// Old (correct):
return arg.required === true ? option : option.optional();

// New (broken):
return option;  // ❌ Always returns string schema, ignoring optional flag
```

**Why This Breaks:**
- Optional arguments will be treated as required
- Validation will fail for commands with optional parameters
- Copilot suggested restoring the previous logic

**Fix Required:**
```typescript
return arg.required === true ? option : option.optional();
```

---

## 📋 **Summary**

### ✅ **Fixed:**
- SonarCloud quality gate passing
- Removed manual stringification logic
- Refactored code structure

### ❌ **Still Broken:**
1. **CRITICAL:** `JSON.stringify()` has no error handling for circular references
2. **HIGH:** Error log level incorrectly set to 'debug'
3. **MEDIUM:** Optional argument validation broken in `argToZod`

### 🔧 **Recommendation:**
Address all three issues above before merging. The `JSON.stringify` circular reference issue is **blocking** as it will cause runtime crashes.

PR #2406, opened 2026-02-06
revert(chore): downgrade tar due to lerna release incompatibility

piyushsinghgaur1:
## Description

This PR reverts the recent `tar` package upgrade that was introduced to address a reported security vulnerability.

Although the newer `tar` version resolves the vulnerability, it is currently incompatible with the **latest version of lerna** used in our release process. When the `tar` version is overridden or upgraded, lerna fails during the release step, preventing successful package publishing.

To unblock releases and maintain stability, this change reverts `tar` to the previously working version until the compatibility issue between lerna and the newer `tar` version is resolved upstream.

## Impact

- Restores successful lerna-based releases
- Temporarily reintroduces the previous `tar` version

## Follow-up

Once the lerna–`tar` compatibility issue is fixed upstream (or an alternative release approach is adopted), we will reapply the `tar` upgrade to address the vulnerability without breaking the release process.

### Links: 
https://github.com/lerna/lerna/pull/4265
https://github.com/lerna/lerna/blob/main/packages/lerna/package.json#L104C1-L105C1


yeshamavani:
Trivy was failing here.. we will resolve that once release is done

Issue #2402, opened 2026-01-29
authentication-service: lastLogin field not updated in JWT token on subsequent logins


muzammilsf:
**Describe the bug**
The lastLogin field in the JWT token is not being updated on subsequent logins. When a user logs in, the token contains the stale/old lastLogin timestamp instead of the current login time.

**To Reproduce**
1. Login as a user (not first-time login) via /auth/login endpoint
2. Exchange the auth code for a token via /auth/token endpoint
3. Decode the resulting accessToken JWT
4. Check the lastLogin claim in the payload
5. The lastLogin value will be from the previous login, not the current one

**Expected behavior**
The lastLogin field in the JWT token should reflect the current login timestamp after each successful login.

**Root Cause**
In src/services/idp-login.service.ts, the generateToken method has this logic (lines 154-161):

if (
  payload.userId &&
  !(await this.userRepo.firstTimeUser(payload.userId))
) {
  await this.userRepo.updateLastLogin(payload.userId);
}

return await this.createJWT(payload, authClient, LoginType.ACCESS);

The issue is twofold:

1. Wrong condition check: The code checks payload.userId, but after login the user data is in payload.user (not payload.userId). The condition fails because payload.userId is undefined, so updateLastLogin() is never called.
2. Stale user object: Even if updateLastLogin() were called, the createJWT method (line 208-209) uses the cached payload.user object:
if (payload.user) {
  user = payload.user;  // ← Uses OLD user object before DB update
}
2. This object still has the old lastLogin value since it's not refetched from the database after the update.

**Suggested Fix**
Update the generateToken method to:
1. Check payload.user?.id instead of payload.userId
2. Update the in-memory payload.user.lastLogin with the current timestamp after the DB update

if (
  payload.user?.id &&
  !(await this.userRepo.firstTimeUser(payload.user.id))
) {
  await this.userRepo.updateLastLogin(payload.user.id);
  payload.user.lastLogin = new Date();  // Update the in-memory object
}

return await this.createJWT(payload, authClient, LoginType.ACCESS);

**Additional context**
- This affects all JWT-based authentication flows using the /auth/token endpoint
- The lastLogin field is important for audit trails, session management, and security monitoring
- Users relying on this field for displaying "last seen" information will see incorrect timestamps

muzammilsf:
Had a conversation with @piyushsinghgaur1 . We found the first issue is version mismatch in our application. So there were two issue related to lastLogin field
[...]

PR #2410, opened 2026-02-09
fix(authentication-service): ensure JWT lastLogin reflects current login time

piyushsinghgaur1:
user lastLogin detail will be a part of the response body 

#2402 
GH-2402

This pull request updates the process for recording and handling the user's last login time across the authentication service. The changes ensure that a consistent timestamp is used when updating the user's last login and that this timestamp is also set in the returned in response body .

Last Login will not be a part of the token instead it will be a part of the response body.


services/authentication-service/src/repositories/user.repository.ts
```diff
@@ -222,11 +222,11 @@ export class UserRepository extends DefaultSoftCrudRepository<
     return user;
   }
 
-  async updateLastLogin(userId: string): Promise<void> {
+  async updateLastLogin(userId: string, time?: number): Promise<void> {
```
samarpan-b, line 225: I disagree to this approach. We should not send the timestamp here. it should always remain Date.now. The issue is this timestamp is not provided into JWT correctly. We should only fix that.

services/authentication-service/src/modules/auth/controllers/login.controller.ts
```diff
@@ -197,7 +197,9 @@ export class LoginController {
         payload.user?.id &&
         !(await this.userRepo.firstTimeUser(payload.user.id))
       ) {
-        await this.userRepo.updateLastLogin(payload.user.id);
+        const time = Date.now();
+        await this.userRepo.updateLastLogin(payload.user.id, time);
```
samarpan-b: use the returned lastlogin timestamp from the method and update it into payload.

services/authentication-service/src/repositories/user.repository.ts
```diff
@@ -222,11 +222,11 @@ export class UserRepository extends DefaultSoftCrudRepository<
...
     await super.updateById(
       userId,
       {
-        lastLogin: Date.now(),
+        lastLogin: time ?? Date.now(),
```
samarpan-b, line 229: return the new lastlogin time from this function

samarpan-b, line 229: return the new lastlogin time from this function

services/authentication-service/src/repositories/user.repository.ts
muzammilsf: This is for consistency that we store the "exact" same timestamp in the database and in the JWT token.

The issue is that if we do Date.now() inside this method and then again after or before calling this method, then the lastLogin field might be off by a few milliseconds.

muzammilsf: @samarpan-b

samarpan-b: I never said to do that twice

samarpan-b (changes requested):

services/authentication-service/src/repositories/user.repository.ts
yeshamavani: Now returning lastLogin as a part of the responseBody

samarpan-b:
Isn't this a breaking change now ?

services/authentication-service/src/services/idp-login.service.ts
```diff
@@ -334,13 +327,21 @@ export class IdpLoginService {
       if (this.userActivity?.markUserActivity)
         this.markUserActivity({...data}, user, userTenant, loginType);
 
+      let lastLogin;
+      if (loginType === LoginType.ACCESS) {
+        lastLogin = Date.now();
+        await this.userRepo.updateLastLogin(user.id ?? '', lastLogin);
+      } else {
+        lastLogin = (await this.userRepo.findById(user.id ?? ''))?.lastLogin;
+      }
```
samarpan-b, line 336: Whats happening here ?

services/authentication-service/src/repositories/user.repository.ts
samarpan-b, line 229: why is time paramter still here ?

services/authentication-service/src/modules/auth/controllers/login.controller.ts
```diff
@@ -193,13 +193,6 @@ export class LoginController {
         user: this.user,
       };
 
-      if (
```
samarpan-b, line 196: why removed ?

services/authentication-service/src/services/idp-login.service.ts
yeshamavani: if its first time login then its the current date and we update that in the DB else if its relogin then we just fetch that from DB 
we dont update the lastlogin time in case of relogin during the same session
relogin as in - generating new token from refresh

services/authentication-service/src/repositories/user.repository.ts
yeshamavani: since we are sending it from our service class - just to keep the exact same date and time in db and response.
await this.userRepo.updateLastLogin(user.id , lastLogin);

services/authentication-service/src/modules/auth/controllers/login.controller.ts
yeshamavani: handled in the service class

PR #2497, opened 2026-04-10
fix(authentication-service): make authorization code one-time-use

### LINKED ISSUE #2495: Authorization Code Can Be Reused Multiple Times in Login Flow (Security Issue)
#### Summary
The current authentication flow allows the authorization code returned by `/auth/login` to be reused multiple times when calling `/auth/token`. This violates expected security behavior, as authorization codes should be single-use tokens.

---

#### Steps to Reproduce
1. Call `POST /auth/login` with valid credentials.
2. Receive an authorization code in the response.
3. Call `POST /auth/token` using the authorization code to obtain an access token.
4. Repeat step 3 using the same authorization code.

---

#### Expected Behavior
- The authorization code should be valid for **one-time use only**.
- Any subsequent attempts to reuse the same code should result in an error (e.g., `invalid_grant` or similar).

---

#### Actual Behavior
- The same authorization code can be reused multiple times to generate new access tokens.
- No restriction or invalidation occurs after the first use.

---

#### Impact
- This creates a potential **security vulnerability**, as intercepted or leaked authorization codes can be reused to obtain multiple access tokens.
- Violates standard practices for secure authentication flows (e.g., OAuth2 authorization code flow).

---

#### Suggested Fix
- Invalidate the authorization code immediately after its first successful use.
- Optionally:
  - Track usage status in the database (e.g., `used: true`).
  - Return an error if a previously used code is presented again.

---

#### Additional Context
Ensuring authorization codes are single-use is critical to prevent replay attacks and align with secure authentication standards.

piyushsinghgaur1:
GH-2495

## Description
This pull request enhances the security of the authorization code flow by enforcing one-time use of authorization codes and refactors time-to-live (TTL) calculations for tokens and keys to use a single class-level constant. The main improvements are grouped below:

**Security Enhancements:**

* Authorization codes are now checked for prior use before processing, and are revoked immediately after successful verification to prevent reuse, enforcing one-time-use as required by the OAuth2 specification.
https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2

**Code Consistency and Refactoring:**

* Introduced a single `msInSecond` class property to consistently handle TTL calculations across methods, replacing repeated inline `ms = 1000` declarations.
* Updated all TTL settings for refresh tokens, access tokens, and public keys to use the new `msInSecond` property for improved maintainability and readability. [[1]](diffhunk://#diff-0af65c9e52326c0d95ae721e7b8b23ae79e97ca07186d43dedafae948128e0c2L226-R241) [[2]](diffhunk://#diff-0af65c9e52326c0d95ae721e7b8b23ae79e97ca07186d43dedafae948128e0c2L326-R340) [[3]](diffhunk://#diff-0af65c9e52326c0d95ae721e7b8b23ae79e97ca07186d43dedafae948128e0c2L614-R627) [[4]](diffhunk://#diff-0af65c9e52326c0d95ae721e7b8b23ae79e97ca07186d43dedafae948128e0c2L640-R655)

**Code Cleanup:**

* Removed redundant local `ms` variable declarations in several methods, relying on the new class property for TTL calculations. [[1]](diffhunk://#diff-0af65c9e52326c0d95ae721e7b8b23ae79e97ca07186d43dedafae948128e0c2L214) [[2]](diffhunk://#diff-0af65c9e52326c0d95ae721e7b8b23ae79e97ca07186d43dedafae948128e0c2L295) [[3]](diffhunk://#diff-0af65c9e52326c0d95ae721e7b8b23ae79e97ca07186d43dedafae948128e0c2L605)

## Before Fix

https://github.com/user-attachments/assets/fc66636b-5a64-4554-b0ac-79ecb5e81832

## After Fix

https://github.com/user-attachments/assets/50b34384-12c1-48b9-affe-0f7015403201

## Type of change

## Checklist:


sandbox/chat-notification-pubnub-example/facade/package.json
```diff
@@ -89,6 +89,8 @@
     },
     "axios": "^1.12.2",
     "minimatch": "^10.2.1",
-    "@isaacs/brace-expansion": "^5.0.1"
+    "@isaacs/brace-expansion": "^5.0.1",
+    "underscore": "^1.13.8",
+    "undici": "^6.24.0"
```
samarpanB, line 94: why this added ?

package.json
```diff
@@ -54,20 +54,18 @@
...
+    "fast-xml-parser": "^5.5.6",
+    "lodash": "^4.18.0",
+    "undici": "^6.24.0",
+    "axios": "1.15.0"
   },
   "dependencies": {
     "@lavamoat/allow-scripts": "^3.4.1",
     "bcrypt": "^5.1.1",
-    "handlebars": "^4.7.8",
+    "handlebars": "^4.7.9",
```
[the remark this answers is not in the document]
samarpanB: I agree

.trivyignore
```diff
@@ -1,4 +1,6 @@
 CVE-2026-23745
 CVE-2026-23950
 CVE-2026-24842
-CVE-2026-26960
\ No newline at end of file
+CVE-2026-26960
+CVE-2026-29786
```
samarpanB: why ignored ? this is not good

services/authentication-service/src/services/idp-login.service.ts
```diff
@@ -165,13 +167,27 @@ export class IdpLoginService {
...
       const payload = (await this.jwtVerifier(resultCode, {
         audience: request.clientId,
       })) as ClientAuthCode<User, typeof User.prototype.id>;
       if (payload.mfa) {
         throw new HttpErrors.Unauthorized(AuthErrorKeys.UserVerificationFailed);
       }
 
+      // Revoke the auth code immediately after verification to prevent reuse
+      await this.revokedTokensRepo.set(
+        resultCode,
```
samarpanB: we should have a new key for this.

samarpanB (changes requested):
test cases update missing

.trivyignore
piyushsinghgaur1: Earlier, these CVEs were temporarily added to `.trivyignore` due to a vulnerability in the `tar` dependency that was being pulled in via `lerna`. At that time, overriding the `tar` version during release was failing due to compatibility issues with the existing `lerna` version.

I have now upgraded `lerna` to a version that resolves this dependency issue, and the vulnerable `tar` version is no longer being used. As a result, these CVEs are no longer applicable and the `.trivyignore` entries can be safely removed.

Going forward, we will avoid suppressing vulnerabilities unless absolutely necessary and ensure proper remediation instead of long-term ignores.

services/authentication-service/src/services/idp-login.service.ts
```diff
@@ -165,13 +167,30 @@ export class IdpLoginService {
...
       const resultCode = await this.codeReader(request.code);
+
+      // Check if the authorization code has already been used (one-time-use enforcement)
+      const authCodeKeyPrefix = 'auth_code';
+      const revokedAuthCodeKey = `${authCodeKeyPrefix}:${resultCode}`;
+      const isCodeRevoked =
+        await this.revokedTokensRepo.get(revokedAuthCodeKey);
+      if (isCodeRevoked) {
+        throw new HttpErrors.Unauthorized(AuthErrorKeys.CodeExpired);
+      }
```
rohit-sourcefuse, line 178: The revocation check here fails open. If `revokedTokensRepo.get(...)` throws — Redis down, network timeout, connection exhaustion — the exception is swallowed by the outer `catch` block and rethrown as `AuthErrorKeys.InvalidCredentials`. The revocation check is silently bypassed, the code is treated as unused, and the request proceeds.

Security controls must fail closed. Wrap the `get` call in its own try-catch:

```typescript
// Check if the authorization code has already been used (one-time-use enforcement)
const authCodeKeyPrefix = 'revoked:auth_code';
const revokedAuthCodeKey = `${authCodeKeyPrefix}:${resultCode}`;

let isCodeRevoked: RevokedToken | null;
try {
  isCodeRevoked = await this.revokedTokensRepo.get(revokedAuthCodeKey);
} catch (repoError) {
  this.logger.error(
    `[AUTH] Revocation store unavailable — denying auth code exchange for safety. Key: ${revokedAuthCodeKey}. Error: ${(repoError as Error).message}`,
  );
  throw new HttpErrors.ServiceUnavailable(
    'Authentication service temporarily unavailable. Please retry.',
  );
}

if (isCodeRevoked) {
  this.logger.warn(
    `[AUTH][SECURITY] Auth code replay attempt detected. Key: ${revokedAuthCodeKey}, clientId: ${request.clientId}`,
  );
  throw new HttpErrors.Unauthorized(AuthErrorKeys.CodeExpired);
}
```

This also adds a security log on replay detection — currently nothing is logged when a reuse attempt occurs, which makes replay attack investigation impossible.

services/authentication-service/src/services/idp-login.service.ts
```diff
@@ -165,13 +167,30 @@ export class IdpLoginService {
...
       if (payload.mfa) {
         throw new HttpErrors.Unauthorized(AuthErrorKeys.UserVerificationFailed);
       }
 
+      // Revoke the auth code immediately after verification to prevent reuse
+      await this.revokedTokensRepo.set(
+        revokedAuthCodeKey,
+        {token: resultCode},
+        {ttl: authClient.authCodeExpiration * this.msInSecond},
+      );
```
rohit-sourcefuse, line 192: **Race condition (TOCTOU) — concurrent replay is not prevented**

Two simultaneous requests carrying the same auth code will both call `revokedTokensRepo.get(...)` and receive `null` (not revoked), both pass `jwtVerifier`, and both reach `revokedTokensRepo.set(...)` — issuing two valid tokens for a single authorization code. This is a realistic attack: an intercepted code replayed in parallel with the legitimate exchange will succeed for both.

The fix requires an **atomic set-if-not-exists** operation — a Redis `SET key value NX EX ttl` that marks the code as used and checks whether it was already used in a single step.

Add this method to `RevokedTokenRepository`:

```typescript
// In services/authentication-service/src/repositories/revoked-token.repository.ts

/**
 * Atomically marks the key as used if not already present.
 * Returns true if the key was set (first use), false if it already existed (replay).
 * Uses Redis SET NX EX for atomicity.
 */
async setIfNotExists(
  key: string,
  value: RevokedToken,
  options: {ttl: number},
): Promise<boolean> {
  // ttl from LoopBack is in milliseconds; Redis EX is in seconds
  const ttlSeconds = Math.ceil(options.ttl / 1000);
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  const connector = (this.dataSource.connector as any);
  const result = await connector.execute(
    'SET',
    [key, JSON.stringify(value), 'NX', 'EX', ttlSeconds],
  );
  // Redis SET NX returns 'OK' on success, null if key already existed
  return result !== null;
}
```

Then replace the current `get` + separate `set` with a single atomic call in `generateToken`:

```typescript
// Replace the isCodeRevoked block AND the later revokedTokensRepo.set block with:
const authCodeKeyPrefix = 'revoked:auth_code';
const revokedAuthCodeKey = `${authCodeKeyPrefix}:${resultCode}`;

let codeWasNew: boolean;
try {
  codeWasNew = await this.revokedTokensRepo.setIfNotExists(
    revokedAuthCodeKey,
    {token: resultCode},
    {ttl: authClient.authCodeExpiration * this.msInSecond},
  );
} catch (repoError) {
  this.logger.error(
    `[AUTH] Revocation store error during auth code exchange. Key: ${revokedAuthCodeKey}. Error: ${(repoError as Error).message}`,
  );
  throw new HttpErrors.ServiceUnavailable(
    'Authentication service temporarily unavailable. Please retry.',
  );
}

if (!codeWasNew) {
  this.logger.warn(
    `[AUTH][SECURITY] Auth code replay attempt. Key: ${revokedAuthCodeKey}, clientId: ${request.clientId}`,
  );
  throw new HttpErrors.Unauthorized(AuthErrorKeys.CodeExpired);
}

// No separate set() call needed — code is already atomically marked above
const payload = (await this.jwtVerifier(resultCode, {
  audience: request.clientId,
})) as ClientAuthCode<User, typeof User.prototype.id>;
```

Note: moving the atomic mark **before** `jwtVerifier` also avoids the window where a JWT verification failure leaves the code permanently revoked without a token ever being issued.

rohit-sourcefuse, line 192: If `revokedTokensRepo.set(...)` throws here — after `jwtVerifier` has already verified the JWT — the outer `catch` block will rethrow it as `AuthErrorKeys.InvalidCredentials` (401). The caller gets an auth failure and will retry with the same code. Because the `set` never succeeded, the code is still not in the revocation store, so the retry passes the revocation check and issues a token.

A transient Redis write failure therefore causes: first attempt → 401, retry → 200 with a token. The auth code has been used without being revoked.

If adopting the atomic `setIfNotExists` approach above, this call is eliminated entirely (the mark happens atomically before `jwtVerifier`). Otherwise, wrap this `set` explicitly:

```typescript
try {
  await this.revokedTokensRepo.set(
    revokedAuthCodeKey,
    {token: resultCode},
    {ttl: authClient.authCodeExpiration * this.msInSecond},
  );
} catch (setError) {
  this.logger.error(
    `[AUTH] Failed to revoke auth code — cannot safely issue token. Key: ${revokedAuthCodeKey}. Error: ${(setError as Error).message}`,
  );
  // Throw 500, not 401 — a 401 causes the client to retry with the same
  // code, which will succeed because revocation was never written.
  throw new HttpErrors.InternalServerError(
    'Token issuance failed due to an internal error. Please restart the authorization flow.',
  );
}
```

services/authentication-service/src/services/idp-login.service.ts
```diff
@@ -165,13 +167,30 @@ export class IdpLoginService {
     }
     try {
       const resultCode = await this.codeReader(request.code);
+
+      // Check if the authorization code has already been used (one-time-use enforcement)
+      const authCodeKeyPrefix = 'auth_code';
+      const revokedAuthCodeKey = `${authCodeKeyPrefix}:${resultCode}`;
```
rohit-sourcefuse: `authCodeKeyPrefix` is a magic string defined inline as a local `const`. The same `revokedTokensRepo` is used by `logoutUser` to revoke access tokens (with no prefix — raw token string as key). There is no central registry of key patterns, so future callers have no way to discover what namespaces exist in this store without scanning every call site.

Extract to a class-level constant:

```typescript
// Add near msInSecond, at the top of the class body:
private readonly msInSecond = 1000;
private static readonly REVOKED_AUTH_CODE_PREFIX = 'revoked:auth_code' as const;
```

Use it as:
```typescript
const revokedAuthCodeKey = `${IdpLoginService.REVOKED_AUTH_CODE_PREFIX}:${resultCode}`;
```

The `revoked:` outer namespace also helps distinguish these keys from access-token revocation entries (which use raw token strings with no prefix), reducing the chance of a collision if other code ever stores keys in the same Redis store.

services/authentication-service/src/__tests__/acceptance/login.controller.acceptance.ts
```diff
@@ -557,6 +557,43 @@ describe('Authentication microservice', () => {
...
+    const replayResponse = await client
+      .post(`/auth/token`)
+      .set(deviceIdName, deviceId)
+      .set(useragentName, useragent)
+      .send({clientId: 'web', code: authCode})
+      .expect(401);
+
+    expect(replayResponse.body.error.message).to.equal(
+      AuthErrorKeys.CodeExpired,
+    );
```
rohit-sourcefuse, line 594: The assertion `expect(replayResponse.body.error.message).to.equal(AuthErrorKeys.CodeExpired)` may be vacuous. Looking at other 401 assertions in this file (e.g. lines ~214, ~373), some access `.body.error.message.message` (nested) rather than `.body.error.message` (flat). If `HttpErrors.Unauthorized` wraps the error key in a nested object, `.body.error.message` resolves to an object (not a string), and `.to.equal(AuthErrorKeys.CodeExpired)` will fail for the wrong reason — or pass vacuously if mocha/chai is not strict about type equality here.

Verify the actual response shape and tighten the assertion:

```typescript
// Log what you actually get, then assert precisely
console.log('replay error body:', JSON.stringify(replayResponse.body.error));

// Once confirmed, use:
expect(replayResponse.body.error).to.not.be.undefined();
expect(replayResponse.body.error.message).to.be.a('string');
expect(replayResponse.body.error.message).to.equal(AuthErrorKeys.CodeExpired);
// Or, if the structure is nested like other tests in this file:
// expect(replayResponse.body.error.message.message).to.equal(AuthErrorKeys.CodeExpired);
```

package.json
```diff
@@ -54,20 +54,18 @@
...
-      "twilio": {
-        "axios": "^1.12.2"
-      }
-    },
     "read-package-json": {
       "glob": "^10.5.0"
     },
-    "fast-xml-parser": "^5.3.6"
+    "fast-xml-parser": "^5.5.6",
+    "lodash": "^4.18.0",
```
rohit-sourcefuse, line 61: `lodash: "^4.18.0"` — lodash 4.18.0 does not appear to exist. The latest stable 4.x release is `4.17.21`. A `^4.18.0` range requires `>=4.18.0 <5.0.0`. If no matching version is in the registry, `npm install` will fail for all contributors with a resolution error.

Change to:
```json
"lodash": "^4.17.21"
```

If `4.18.0` was recently published and this is intentional, please add a comment with the CVE or reason this specific version is needed.

package.json
```diff
@@ -54,20 +54,18 @@
...
+    "undici": "^6.24.0",
+    "axios": "1.15.0"
```
rohit-sourcefuse, line 63: `"axios": "1.15.0"` pins to an exact version with a flat root override, replacing the previous nested override that scoped the axios upgrade specifically to `loopback4-notifications > twilio`. The flat override forces every package in the monorepo that depends on axios — regardless of its stated range — to use exactly `1.15.0`. If any future dependency requires a newer patch (e.g. a security fix in `1.15.x`), it will be silently blocked.

Use a range instead:
```json
"axios": "^1.15.0"
```

If the intent was only to patch the twilio transitive path, consider restoring the nested structure to limit blast radius:
```json
"loopback4-notifications": {
  "twilio": {
    "axios": "^1.15.0"
  }
}
```

.github/workflows/trivy.yml
```diff
@@ -22,8 +22,8 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Run Trivy vulnerability scanner in repo mode
-        uses: aquasecurity/trivy-action@0.28.0
+        uses: aquasecurity/trivy-action@0.35.0
```
rohit-sourcefuse: `aquasecurity/trivy-action@0.35.0` is a mutable tag — same concern flagged across other repos in this org. The upstream maintainer can force-push this tag to a different commit at any time.

Also, note that the SHA `57a97c7e7821a5776cebc9bb87c984fa69cba8f1` referenced in earlier PR comments **does not resolve** — CI confirmed this with the error:
> `Unable to resolve action aquasecurity/trivy-action@57a97c7e7821a5776cebc9bb87c984fa69cba8f1`

To get the correct SHA for `v0.35.0`:
```bash
git ls-remote https://github.com/aquasecurity/trivy-action refs/tags/v0.35.0
# Use the dereferenced commit (the ^{} line), not the tag object SHA
```

Then pin:
```yaml
uses: aquasecurity/trivy-action@<correct-sha-here> # v0.35.0
```

rohit-sourcefuse (changes requested):
The security goal here is correct and the OAuth2 one-time-use enforcement is a real requirement (RFC 6749 §4.1.2). The `msInSecond` refactor is also a clean, worthwhile change. That said, the implementation of the core security feature has two critical correctness flaws that defeat the guarantee under realistic conditions, plus a few package-level issues that need attention before this can merge.

---

### Must fix before merge

**1. Race condition (TOCTOU) — inline comments**
Two concurrent requests with the same auth code both pass the `revokedTokensRepo.get(...)` check before either has written the revocation entry. The fix requires an atomic `SET NX EX` operation (Redis "set if not exists") on `RevokedTokenRepository`. Full working code for both the repository method and the `generateToken` call site is in the inline comment on line 192.

**2. Fail-open on Redis error — inline comment on line 178**
If `revokedTokensRepo.get(...)` throws, the outer catch swallows it and the request proceeds as if the code was unused. Security controls must fail closed — wrap the `get` call separately and throw a 503 if the revocation store is unreachable.

**3. `revokedTokensRepo.set` failure causes retryable code reuse — inline comment on line 192**
If the `set` throws after `jwtVerifier` succeeds, the outer catch rethrows as a 401. The client retries with the same (never-revoked) code and succeeds. Addressed by the atomic approach above, or by wrapping the `set` and throwing a 500 explicitly. A 500 signals "restart the flow"; a 401 signals "retry the same request."

**4. `lodash: "^4.18.0"` override breaks `npm install` — inline comment on line 61 of `package.json`**
Lodash 4.18.0 does not exist; the latest 4.x is 4.17.21. A `^4.18.0` range will fail to resolve. Change to `^4.17.21`.

**5. `package-lock.json` — 4,025 `integrity` hashes removed**
This PR strips every `integrity` SHA-512 and `resolved` URL from `package-lock.json`. These are what `npm ci` uses to verify downloaded tarballs match expected content. Without them, the install pipeline fetches packages without verifying their content — a supply-chain regression that is particularly ironic in a security-hardening PR. This looks like an artifact of regenerating the lockfile with a different npm version. Please revert to a lockfile that includes integrity hashes, or explain in the PR why stripping them is intentional and safe.

---

### .trivyignore deletion

The file was suppressing 4 CVEs (`CVE-2026-23745`, `CVE-2026-23950`, `CVE-2026-24842`, `CVE-2026-26960`). Please add a note in the description confirming which dependency version bump in this PR addresses each CVE — or re-add the entries with inline comments if they are false positives. Without that context, reviewers cannot verify whether Trivy CI will pass or what the actual exposure is.

---

### Nice-to-have (follow-up)

- Add unit tests for `generateToken` with a mocked `revokedTokensRepo` covering the store-failure path and, once the atomic approach is in place, the concurrent-request scenario
- Extract `authCodeKeyPrefix` to a `private static readonly` class constant (inline comment on line 173)
- Verify the error message assertion in the new acceptance test against the actual response body structure (inline comment on line 594)
- Pin `trivy-action` to an immutable commit SHA (inline comment on `.github/workflows/trivy.yml` line 25 — note that the SHA previously suggested in this repo's PR comments does not resolve; run `git ls-remote` to get the correct one for `v0.35.0`)

services/authentication-service/src/services/idp-login.service.ts
piyushsinghgaur1: Now using auth_code:<code> instead of the raw code to avoid mixing with other revoked tokens and to keep key space isolated.

sandbox/chat-notification-pubnub-example/facade/package.json
piyushsinghgaur1: trivy vulnerability fix

package.json
piyushsinghgaur1: It is available.

image

rohit-sourcefuse: Thanks for the screenshot — confirmed `4.18.0` is published. Withdrawing that concern, the override is valid.

services/authentication-service/src/services/idp-login.service.ts
rohit-sourcefuse: The `auth_code:` prefix is already in the current code — the concern was that it's defined as a local `const` inside the method rather than a `private static readonly` class constant. As-is, if any future code needs to look up or delete these keys, they'd have to duplicate the string or hunt through the method to find it.

Small change — just move it to the class body:
```typescript
private static readonly REVOKED_AUTH_CODE_PREFIX = 'auth_code' as const;
```
Then use `IdpLoginService.REVOKED_AUTH_CODE_PREFIX` at the call site. The key format itself is fine.

.trivyignore
rohit-sourcefuse: The explanation makes sense — `lerna` upgrade resolves the `tar` vulnerability, so the CVE suppressions are no longer needed. Good to remove them properly rather than carry them indefinitely.

package.json
piyushsinghgaur1: Added "axios": "1.15.0" intentionally as a strict pin due to the recent security concern around axios. The goal was to avoid any unintended upgrades/downgrades and ensure we are locked to a verified safe version across the monorepo.

services/authentication-service/src/services/idp-login.service.ts
```diff
@@ -165,13 +167,27 @@ export class IdpLoginService {
     }
     try {
       const resultCode = await this.codeReader(request.code);
+
+      // Check if the authorization code has already been used (one-time-use enforcement)
+      const isCodeRevoked = await this.revokedTokensRepo.get(resultCode);
+      if (isCodeRevoked) {
+        throw new HttpErrors.Unauthorized(AuthErrorKeys.CodeExpired);
+      }
```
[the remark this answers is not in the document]
piyushsinghgaur1: Added

rohit-sourcefuse (approved):

services/authentication-service/src/services/idp-login.service.ts
piyushsinghgaur1: fixed

.github/workflows/trivy.yml
piyushsinghgaur1: fixed

yeshamavani (approved):

PR #2502, opened 2026-04-13
feat(observability): add @sourceloop/observability with OTEL-first bootstrap, profiles, and pluggable instrumentation

samarpan-b:
## Description

This PR introduces a new reusable package, @sourceloop/observability, to provide a generic OpenTelemetry-first observability layer for Sourceloop services.

The package is designed around:

- early bootstrap for tracing setup before app startup
- optional LoopBack integration through ObservabilityComponent
- profile-based configuration for OTLP, New Relic, SigNoz, and Datadog
- optional instrumentation packages that consumers can install and enable selectively
- generic tracing helpers for spans, errors, and propagation

## Configuration model

Startup-critical observability config is resolved during bootstrap from:

1. explicit bootstrap overrides
2. environment variables
3. defaults

LoopBack DI config is still supported, but only for component-level enrichment rather than early startup control.

## Instrumentation behavior

By default, only:

- http
- express

are enabled.

All other built-in instrumentations are disabled by default and must be explicitly enabled through bootstrap config or env vars.

Instrumentation packages are optional peer dependencies. If an instrumentation is enabled but its package is not installed, bootstrap fails fast with a clear validation error.

## Type of change

- [x] Intermediate change (work in progress)

## How Has This Been Tested?

Verified locally with:

- npm run build
- npm test
- npm run lint

## Checklist:


packages/observability/src/profiles/newrelic.profile.ts
```diff
@@ -0,0 +1,50 @@
...
+      },
+      resourceAttributes: {
+        'vendor.apm': 'newrelic',
+        ...config.resourceAttributes,
+      },
+    };
+  }
+
+  initialize(
+    config: ResolvedObservabilityConfig,
```
rohit-sourcefuse: I noticed that `bootstrapObservability()` in `bootstrap.ts` already calls `profile.applyDefaults(resolvedConfig)` before invoking `profile.initialize()`. Since this override calls `this.applyDefaults(config)` again before delegating to `super`, `applyDefaults` ends up running twice — headers get merged twice and `vendor.apm` gets set twice. The same pattern exists in `DatadogObservabilityProfile` and `SignozObservabilityProfile`.

Removing the `initialize()` override here (and in Datadog/Signoz) should be enough — `BaseOtlpObservabilityProfile.initialize` handles initialization fully, and `applyDefaults` is already called by the bootstrap flow.

```typescript
// applyDefaults() is already called by bootstrapObservability() before initialize() is invoked.
// The base class implementation is sufficient — no need to override initialize() here.
export class NewRelicObservabilityProfile extends BaseOtlpObservabilityProfile {
  name: ObservabilityProfileName = 'newrelic';

  applyDefaults(config: ResolvedObservabilityConfig): ResolvedObservabilityConfig {
    const licenseKey = process.env.NEW_RELIC_LICENSE_KEY?.trim();
    return {
      ...config,
      otlpEndpoint:
        config.otlpEndpoint ??
        (config.exporterProtocol === 'grpc'
          ? 'https://otlp.nr-data.net:4317'
          : 'https://otlp.nr-data.net:4318/v1/traces'),
      otlpHeaders: {
        ...(licenseKey ? {'api-key': licenseKey} : {}),
        ...config.otlpHeaders,
      },
      resourceAttributes: {
        'vendor.apm': 'newrelic',
        ...config.resourceAttributes,
      },
    };
  }
}
```

With the double call, the `config.otlpHeaders` spread runs twice, which can put headers in a state that's tricky to trace back to the source.

packages/observability/src/tracing.ts
```diff
@@ -0,0 +1,38 @@
...
+  return getTracer().startActiveSpan(name, async span => {
+    if (attributes) {
+      span.setAttributes(attributes);
+    }
+
+    try {
+      return await fn();
+    } catch (error) {
+      span.recordException(error as Error);
+      throw error;
```
rohit-sourcefuse, line 26: The catch block records the exception but doesn't follow up with `span.setStatus({ code: SpanStatusCode.ERROR })`. Per the OpenTelemetry spec, recording an exception and setting the span status are two separate operations — the exception is attached as an event on the span, but the span status stays `UNSET` unless explicitly updated. Most trace backends (Jaeger, Grafana Tempo, Datadog, New Relic) use the span status to colour and surface error traces, so without this, spans that throw would appear successful in dashboards.

Worth noting that `recordException` in `errors.ts` already handles this correctly — `withSpan` just needs to match it:

```typescript
import {context, trace, SpanStatusCode} from '@opentelemetry/api';

// inside the catch block:
} catch (error) {
  span.recordException(error as Error);
  span.setStatus({
    code: SpanStatusCode.ERROR,
    message: error instanceof Error ? error.message : String(error),
  });
  throw error;
} finally {
  span.end();
}
```

This is the one thing I'd consider a hard blocker — without it the package would make errors harder to find, which is the opposite of what observability is for.

packages/observability/src/bootstrap.ts
```diff
@@ -0,0 +1,124 @@
...
+    updateRuntimeState({runtime});
+    return runtime;
+  }
+
+  const profile = resolveProfile(resolvedConfig.profile);
+  const profileConfig = profile.applyDefaults(resolvedConfig);
+  validateObservabilityConfig(profileConfig);
+
+  const initResult = profile.initialize(profileConfig, {
+    createExporter: createOtlpExporter,
```
rohit-sourcefuse: Two related reliability gaps I wanted to flag together here.

First — if `profile.initialize()` throws (an instrumentation ABI mismatch, a `tracerProvider.register()` failure, a missing peer dep slipping past validation), `updateRuntimeState({profile, config})` has already run but `updateRuntimeState({runtime})` hasn't. A subsequent call to `bootstrapObservability()` then sees `runtime === undefined`, re-enters against corrupted partial state, and the original error is never logged. This can be surprisingly hard to diagnose in production.

Second — `await runtime.shutdown()` in `shutdownObservability()` isn't wrapped in a try-catch. The OTel SDK's `tracerProvider.shutdown()` can reject on flush timeout, which becomes an unhandled rejection in a SIGTERM handler and crashes the process in Node 15+.

```typescript
// bootstrapObservability — wrap initialization in try-catch
try {
  const profile = resolveProfile(resolvedConfig.profile);
  const profileConfig = profile.applyDefaults(resolvedConfig);
  validateObservabilityConfig(profileConfig);

  const initResult = profile.initialize(profileConfig, {
    createExporter: createOtlpExporter,
  });

  const runtime: ObservabilityRuntime = {
    enabled: true,
    profile: profile.name,
    config: profileConfig,
    tracerProvider: initResult.tracerProvider,
    async shutdown() {
      await profile.shutdown?.();
      clearRuntimeState();
    },
  };

  updateRuntimeState({profile, config: profileConfig, runtime});
  return runtime;
} catch (err) {
  clearRuntimeState(); // prevent corrupted partial state on retry
  console.error('[observability] Initialization failed. Tracing will be disabled.', err);
  throw err;
}

// shutdownObservability — guard against rejection
export async function shutdownObservability(): Promise<void> {
  const runtime = getRuntimeState().runtime;
  if (!runtime) return;

  try {
    await runtime.shutdown();
  } catch (err) {
    // Log but don't rethrow — a shutdown error must not crash
    // the process during SIGTERM handling.
    console.error(
      '[observability] Error during shutdown. Some spans may not have been flushed.',
      err,
    );
  }
}
```

The `clearRuntimeState()` on failure is the key part — it ensures a retry starts clean rather than against partial state.

rohit-sourcefuse (changes requested):
Really impressive work here — the 3-tier config pipeline, the `BaseOtlpObservabilityProfile` abstraction shared cleanly across all four vendor profiles, and the lazy `require()` approach for optional instrumentation are all the right design choices for a package like this. The `ProfileBootstrapContext` injection seam is a particularly nice touch that makes testing straightforward without pulling in real OTel packages.

A few things I've flagged inline that I'd like to see addressed before this merges:

**Worth resolving before merge:**
- Double `applyDefaults` in `NewRelicObservabilityProfile`, `DatadogObservabilityProfile`, and `SignozObservabilityProfile` — `bootstrap.ts` already calls `applyDefaults` before `initialize`, so the overrides run it twice
- `SpanStatusCode.ERROR` missing in `withSpan` catch block — without this, error spans appear successful in every trace backend (Jaeger, Datadog, Grafana Tempo, etc.)
- No error handling around `profile.initialize()` in `bootstrapObservability()` — a failure there leaves the module singleton in a corrupted partial state with no logging; and `shutdownObservability()` needs a try-catch guard since `tracerProvider.shutdown()` can reject in a SIGTERM handler

**Happy to defer to a follow-up:**
- `ObservabilityProfileRegistry` is wired into the component but `bootstrapObservability()` never consults it — a custom profile registered via LoopBack DI is silently ignored today. Worth either connecting it or leaving a comment noting it's planned for a later iteration
- Exporter peer deps (`@opentelemetry/exporter-trace-otlp-http`, `-grpc`) don't have the same pre-flight check that instrumentation packages get in `validateObservabilityConfig` — a missing exporter gives a raw `MODULE_NOT_FOUND` instead of a helpful install prompt
- Test coverage for the idempotency guard, `shutdownObservability()`, and the `withSpan` error rethrow path

Also worth confirming — are the prettier failures and Node 24 EBADENGINE warnings in CI pre-existing on the base branch, or introduced here?

packages/observability/src/component.ts
```diff
@@ -0,0 +1,56 @@
...
+export class ObservabilityComponent implements Component {
+  bindings: Binding[] = [];
+
+  constructor(
+    @inject(CoreBindings.APPLICATION_INSTANCE)
+    private readonly application: RestApplication,
+    @inject(ObservabilityBindings.config, {optional: true})
+    private readonly config?: ObservabilityConfig,
+  ) {
+    const runtime = getRuntimeState().runtime ?? bootstrapObservability();
```
sf-sahil-jassal: 🟠 **Architecture: No LoopBack Sequence integration for tracer/meter enrichment**

**Issue:** The component bootstraps OTEL but never wires tracer or meter into the LoopBack request sequence. Without a custom `Sequence` (or `SequenceActions`), there is no automatic span-per-request, no correlation of `traceId`/`spanId` into request logs, and no request-level metrics (counters, histograms). The `tracerProvider` sits in the DI container but nothing starts a root span for incoming HTTP requests.

**Fix:** Add an `ObservabilitySequence` (or `ObservabilityAction`) that wraps each request in a span and injects trace context into the logger:

```typescript
// packages/observability/src/sequence.ts
import {inject} from '@loopback/core';
import {RestBindings, SequenceHandler, RequestContext} from '@loopback/rest';
import {trace, context, propagation} from '@opentelemetry/api';
import {ObservabilityBindings} from './keys';

export class ObservabilitySequence implements SequenceHandler {
  constructor(
    @inject(RestBindings.Http.CONTEXT)
    private readonly requestCtx: RequestContext,
    @inject(ObservabilityBindings.runtime, {optional: true})
    private readonly runtime?: ObservabilityRuntime,
  ) {}

  async handle(request: Request, response: Response): Promise<void> {
    const tracer = trace.getTracer('@sourceloop/observability');
    const carrier: Record<string, string> = {};
    // Extract incoming W3C traceparent from request headers
    propagation.extract(context.active(), request.headers, {
      get: (h, k) => h[k] as string,
      keys: h => Object.keys(h),
    });
    // Start root span per request
    await tracer.startActiveSpan(`${request.method} ${request.path}`, async span => {
      try {
        // inject traceId into logger context here
        await next(); // delegate to LoopBack routing
      } catch (err) {
        span.recordException(err as Error);
        throw err;
      } finally {
        span.end();
      }
    });
  }
}
```

Then bind it in `ObservabilityComponent`:
```typescript
this.application.sequence(ObservabilitySequence);
```

This is the standard LoopBack 4 pattern for cross-cutting request concerns (auth, logging, tracing).

**Credit:** Open Code

packages/observability/src/types.ts
```diff
@@ -0,0 +1,100 @@
...
+  otlpHeaders: Record<string, string>;
+  exporterProtocol: ExporterProtocol;
+  sampler: SamplerName;
+  samplerArg: number;
+  instrumentations: InstrumentationToggles;
+  customInstrumentations: ObservabilityInstrumentation[];
+  resourceAttributes: Record<string, string>;
+}
+
+export interface ProfileInitResult {
```
sf-sahil-jassal, line 66: 🟠 **Missing: Metrics (MeterProvider) support — traces-only design**

**Issue:** `ProfileInitResult` and `ObservabilityRuntime` expose only `tracerProvider`. There is no `meterProvider` in the type definitions, no `@opentelemetry/sdk-metrics` integration, and no way for consumers to record counters, histograms, or gauges. The PR title says "OTEL-first" but ships traces only — metrics and logs are absent from the type contract.

**Fix:** Extend the types to include metrics:

```typescript
import {MeterProvider} from '@opentelemetry/sdk-metrics';

export interface ProfileInitResult {
  exporterName: string;
  tracerProvider: NodeTracerProvider;
  meterProvider?: MeterProvider; // add
}

export interface ObservabilityRuntime {
  enabled: boolean;
  profile: ObservabilityProfileName;
  config: ResolvedObservabilityConfig;
  tracerProvider?: NodeTracerProvider;
  meterProvider?: MeterProvider; // add
  shutdown(): Promise<void>;
}
```

And wire `MeterProvider` in `BaseOtlpObservabilityProfile.initialize()` alongside `NodeTracerProvider`.

**Credit:** Open Code

packages/observability/src/runtime.ts
```diff
@@ -0,0 +1,33 @@
...
+} from './types';
+
+export type RuntimeState = {
+  bootstrapOverrides?: Partial<ObservabilityConfig>;
+  config?: ResolvedObservabilityConfig;
+  profile?: ObservabilityProfile;
+  runtime?: ObservabilityRuntime;
+};
+
+const state: RuntimeState = {};
```
sf-sahil-jassal, line 15: 🔴 **Concurrency: Module-level mutable singleton breaks test isolation and multi-instance apps**

**Issue:** `const state: RuntimeState = {}` is a module-level singleton mutated in place. In test suites that don't call `clearRuntimeState()` between tests (or where tests run in parallel), state bleeds across tests. In monorepo scenarios where multiple services are loaded in the same process (e.g. via `require`), they share this singleton — whichever bootstraps first wins and silently ignores subsequent configs.

**Fix:** Either:
1. Use a `Symbol`-keyed property on `globalThis` so it is truly process-global and explicit, or
2. Accept a `store` parameter in `getRuntimeState`/`updateRuntimeState` for testability:

```typescript
// Explicit global registry (option 1)
const STATE_KEY = Symbol.for('@sourceloop/observability.state');
function getGlobalState(): RuntimeState {
  if (!(globalThis as Record<symbol, unknown>)[STATE_KEY]) {
    (globalThis as Record<symbol, unknown>)[STATE_KEY] = {};
  }
  return (globalThis as Record<symbol, unknown>)[STATE_KEY] as RuntimeState;
}
```

Also: `clearRuntimeState` uses property-by-property `= undefined` which is fragile — add a new key to `RuntimeState` and you must remember to clear it here too. Prefer `Object.assign(state, {bootstrapOverrides: undefined, config: undefined, profile: undefined, runtime: undefined})` or delete all keys in a loop.

**Credit:** Open Code

packages/observability/src/bootstrap.ts
```diff
@@ -0,0 +1,124 @@
...
+        enabled: false,
+        profile: 'none',
+      }),
+    async shutdown() {
+      clearRuntimeState();
+    },
+  };
+}
+
+export function bootstrapObservability(
```
sf-sahil-jassal, line 64: 🟠 **No mutex / race guard on concurrent `bootstrapObservability()` calls**

**Issue:** `bootstrapObservability` checks `getRuntimeState().runtime` and returns early if set. But two concurrent callers (e.g., two services starting simultaneously in the same process) can both pass the guard before either sets `updateRuntimeState({runtime})` at line 101. This results in double-registration of the `tracerProvider` and two sets of instrumentations being activated.

**Fix:** Add a synchronous initialization lock:

```typescript
let bootstrapping = false;

export function bootstrapObservability(
  overrides?: Partial<ObservabilityConfig>,
): ObservabilityRuntime {
  const existingRuntime = getRuntimeState().runtime;
  if (existingRuntime) return existingRuntime;
  if (bootstrapping) {
    throw new Error('bootstrapObservability called re-entrantly; ensure it is called once at startup.');
  }
  bootstrapping = true;
  try {
    // ... existing logic
  } finally {
    bootstrapping = false;
  }
}
```

**Credit:** Open Code

packages/observability/src/profiles/otlp.profile.ts
```diff
@@ -0,0 +1,141 @@
...
+  if (config.environment) {
+    attributes[ATTR_DEPLOYMENT_ENVIRONMENT] = config.environment;
+  }
+
+  return detectResourcesSync().merge(new Resource(attributes));
+}
+
+function loadOtlpExporter(config: ResolvedObservabilityConfig): SpanExporter {
+  if (config.exporterProtocol === 'grpc') {
+    const otlpGrpc = require('@opentelemetry/exporter-trace-otlp-grpc') as {
```
sf-sahil-jassal: 🟠 **`require()` inside methods bypasses module bundlers and breaks type safety**

**Issue:** `loadOtlpExporter` uses `require('@opentelemetry/exporter-trace-otlp-grpc')` and `require('@opentelemetry/exporter-trace-otlp-http')` inside the function body. This pattern:
- Prevents tree-shaking and static analysis
- Gives `any`-typed return that is immediately cast — if the module shape changes, this fails silently at runtime
- Errors thrown during `require()` are uncaught here and will crash the process with an unhelpful stack trace

**Fix:** Wrap in try/catch with a descriptive error, or use dynamic `import()` (which is async and works with ESM):

```typescript
function loadOtlpExporter(config: ResolvedObservabilityConfig): SpanExporter {
  try {
    if (config.exporterProtocol === 'grpc') {
      // eslint-disable-next-line @typescript-eslint/no-var-requires
      const {OTLPTraceExporter} = require('@opentelemetry/exporter-trace-otlp-grpc');
      return new OTLPTraceExporter({url: config.otlpEndpoint, metadata: ...});
    }
    const {OTLPTraceExporter} = require('@opentelemetry/exporter-trace-otlp-http');
    return new OTLPTraceExporter({url: config.otlpEndpoint, headers: ...});
  } catch (err) {
    throw new Error(
      `Failed to load OTLP exporter for protocol "${config.exporterProtocol}". ` +
      `Install the required peer dependency. Original error: ${(err as Error).message}`
    );
  }
}
```

**Credit:** Open Code

packages/observability/src/correlation.ts
```diff
@@ -0,0 +1,26 @@
+import {context, propagation, trace} from '@opentelemetry/api';
+
+export function createPropagationHeaders(
+  carrier: Record<string, string> = {},
+): Record<string, string> {
+  propagation.inject(context.active(), carrier);
+  return carrier;
+}
+
+export function getTraceContext(): {
```
sf-sahil-jassal, line 10: 🟡 **`getTraceContext` always calls `createPropagationHeaders()` — allocates a new carrier even when no span is active**

**Issue:** `getTraceContext()` calls `createPropagationHeaders()` unconditionally, which calls `propagation.inject(context.active(), carrier)`. When no span is active (`activeSpan` is undefined), this allocates and injects into an empty carrier unnecessarily on every call.

Also, for enriching logs/metrics with trace correlation, consumers typically need `traceId` and `spanId` in a structured format, but the returned object mixes OTEL-internal IDs with W3C propagation headers — two different concerns.

**Fix:** Guard the propagation call:

```typescript
export function getTraceContext() {
  const activeSpan = trace.getSpan(context.active());
  const spanContext = activeSpan?.spanContext();
  if (!spanContext) {
    return {traceId: undefined, spanId: undefined, traceparent: undefined, tracestate: undefined};
  }
  const carrier = createPropagationHeaders();
  return {
    traceId: spanContext.traceId,
    spanId: spanContext.spanId,
    traceparent: carrier.traceparent,
    tracestate: carrier.tracestate,
  };
}
```

**Credit:** Open Code

packages/observability/src/config/validate-config.ts
```diff
@@ -0,0 +1,83 @@
+import {ResolvedObservabilityConfig} from '../types';
+import {
+  getEnabledInstrumentations,
+  InstrumentationModuleRequirement,
+  InstrumentationName,
+  INSTRUMENTATION_MODULE_REQUIREMENTS,
+} from '../profiles/instrumentations';
+
+function assertDependencyInstalled(moduleName: string, message: string): void {
```
sf-sahil-jassal, line 10: 🟡 **`isDependencyInstalled` duplicates `isModuleInstalled` in `instrumentations.ts`**

**Issue:** `isDependencyInstalled` (line 17) and `assertDependencyInstalled` (line 9) both use `require.resolve` inside try/catch — the exact same pattern as `isModuleInstalled` in `profiles/instrumentations.ts:45`. This is code duplication across two files.

**Fix:** Extract a shared utility into `packages/observability/src/utils/module-resolution.ts`:

```typescript
export function isModuleInstalled(moduleName: string): boolean {
  try { require.resolve(moduleName); return true; } catch { return false; }
}

export function assertModuleInstalled(moduleName: string, message: string): void {
  if (!isModuleInstalled(moduleName)) throw new Error(message);
}
```

Import from both `validate-config.ts` and `instrumentations.ts`.

**Credit:** Open Code

packages/observability/src/profiles/otlp.profile.ts
```diff
@@ -0,0 +1,141 @@
...
+      sampler: resolveSampler(config),
+      spanProcessors: [new BatchSpanProcessor(exporter)],
+    });
+
+    tracerProvider.register();
+    this.instrumentations = createAutoInstrumentations(config, tracerProvider);
+    this.tracerProvider = tracerProvider;
+
+    return {
+      exporterName: loadOtlpExporter.name,
```
sf-sahil-jassal: 🟡 **`exporterName` returns function name of internal helper — leaks implementation detail**

**Issue:** `ProfileInitResult.exporterName` is set to `loadOtlpExporter.name` (line 117), which evaluates to the string `'loadOtlpExporter'`. This exposes an internal private function name in the public API contract. The integration test at `runtime.integration.ts:85` asserts `strictEqual(initResult.exporterName, 'loadOtlpExporter')` — this will silently break if the function is renamed or minified.

**Fix:** Use a stable semantic string:

```typescript
return {
  exporterName: `otlp-${config.exporterProtocol}`, // e.g. 'otlp-grpc' or 'otlp-http/protobuf'
  tracerProvider,
};
```

Update the integration test assertion accordingly.

**Credit:** Open Code

packages/observability/src/component.ts
sf-sahil-jassal: 🟠 **`bootstrapObservability()` called in component constructor — side effects at DI bind time**

**Issue:** Line 40 calls `getRuntimeState().runtime ?? bootstrapObservability()`. `bootstrapObservability()` registers the global `NodeTracerProvider` (via `tracerProvider.register()` in `otlp.profile.ts:112`), which patches `@opentelemetry/api` globals. This happens during LoopBack DI binding resolution, before the app is fully started. If the app fails to start after this point, the global tracer provider is registered but never shut down, leaking instrumentation hooks.

**Fix:** Move `bootstrapObservability()` to a LoopBack lifecycle hook (`start`/`stop`) to align with the application lifecycle:

```typescript
export class ObservabilityComponent implements Component, LifeCycleObserver {
  async start(): Promise<void> {
    const runtime = bootstrapObservability(this.config);
    this.application.bind(ObservabilityBindings.runtime).to(runtime);
  }

  async stop(): Promise<void> {
    await shutdownObservability();
  }
}
```

**Credit:** Open Code

packages/observability/src/profiles/instrumentations.ts
```diff
@@ -0,0 +1,153 @@
...
+function isModuleInstalled(moduleName: string): boolean {
+  try {
+    require.resolve(moduleName);
+    return true;
+  } catch {
+    return false;
+  }
+}
+
+function loadInstrumentations(
```
sf-sahil-jassal, line 54: 🟠 **Maintainability: Use `@opentelemetry/auto-instrumentations-node` instead of individual instrumentation packages**
**Issue:** The current implementation manually `require()`s 6+ individual OTEL instrumentation packages and maintains a custom `loadInstrumentations` switch-case registry. Problems:
- **Maintenance burden:** Every new library requires a new `case`, a new `INSTRUMENTATION_MODULE_REQUIREMENTS` entry, and a new `InstrumentationToggles` key
- **Version drift:** 6+ packages each have independent release cycles — keeping them in sync with the SDK is error-prone
- **Incomplete coverage:** Misses many libraries the OTEL ecosystem already supports (MongoDB, TypeORM, gRPC, DNS, Net, etc.)
**Fix:** Replace the entire `loadInstrumentations` switch with `@opentelemetry/auto-instrumentations-node`, which bundles all supported instrumentations and accepts a per-package `{enabled: boolean}` config:

package.json — one dep instead of 6+:
```json
{
  "peerDependencies": {
    "@opentelemetry/auto-instrumentations-node": "^0.56.0"
  }
}
```

```typescript
import {getNodeAutoInstrumentations} from '@opentelemetry/auto-instrumentations-node';
import {NodeTracerProvider} from '@opentelemetry/sdk-trace-node';
import {InstrumentationToggles, ResolvedObservabilityConfig} from '../types';
// Maps our toggle keys to the package names auto-instrumentations-node expects
const TOGGLE_TO_PACKAGE: Record<keyof InstrumentationToggles, string[]> = {
  http:    ['@opentelemetry/instrumentation-http'],
  express: ['@opentelemetry/instrumentation-express'],
  pg:      ['@opentelemetry/instrumentation-pg'],
  mysql:   ['@opentelemetry/instrumentation-mysql', '@opentelemetry/instrumentation-mysql2'],
  redis:   ['@opentelemetry/instrumentation-redis-4'],
  kafka:   ['@opentelemetry/instrumentation-kafkajs'],
};
export function createAutoInstrumentations(
  config: ResolvedObservabilityConfig,
  tracerProvider: NodeTracerProvider,
) {
  const enabledPackages = new Set<string>();
  for (const [toggle, packages] of Object.entries(TOGGLE_TO_PACKAGE)) {
    if (config.instrumentations[toggle as keyof InstrumentationToggles]) {
      packages.forEach(p => enabledPackages.add(p));
    }
  }
  // All packages disabled by default; only toggled-on ones are active
  const allPkgs = Object.values(TOGGLE_TO_PACKAGE).flat();
  const instrumentations = getNodeAutoInstrumentations(
    Object.fromEntries(allPkgs.map(pkg => [pkg, {enabled: enabledPackages.has(pkg)}]))
  );
  const all = [...instrumentations, ...config.customInstrumentations];
  for (const inst of all) {
    inst.setTracerProvider?.(tracerProvider);
    inst.enable();
  }
  return all;
}
```


Benefits:
- Single peer dep to maintain and version-bump
- Automatically gains new instrumentations as the package adds them
- enabled: false entries are zero-cost no-ops at runtime
- Backwards-compatible with existing InstrumentationToggles contract — no breaking change for consumers
- Eliminates INSTRUMENTATION_MODULE_REQUIREMENTS and the peer dep checks in validate-config.ts entirely — auto-instrumentations-node handles graceful skipping internally

**Credit**: Open Code

sf-sahil-jassal (changes requested):

packages/observability/src/tracing.ts
samarpan-b: Fixed. withSpan() now marks failed spans with SpanStatusCode.ERROR in addition to recording the exception, so traces no longer appear successful on error paths.

packages/observability/src/bootstrap.ts
samarpan-b: Fixed. bootstrap now guards profile initialization and clears partial runtime state on failure. shutdownObservability() also handles shutdown errors defensively so process termination does not leave the singleton in a corrupt state.

packages/observability/src/profiles/newrelic.profile.ts
samarpan-b: Fixed. profile defaults are now applied once during bootstrap resolution; vendor profile initialize() no longer reapplies them.

packages/observability/src/runtime.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function getRuntimeState(): RuntimeState {
+  return state;
+}
```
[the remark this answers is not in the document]
samarpan-b: Good catch. I changed getRuntimeState() to expose a safe readonly/snapshot view instead of the mutable backing object.

packages/observability/src/profiles/otlp.profile.ts
```diff
@@ -0,0 +1,141 @@
...
+      tracerProvider,
+    };
```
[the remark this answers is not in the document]
samarpan-b: Fixed. exporter metadata now reflects the actual exporter instance/context instead of assuming the built-in OTLP loader.

packages/observability/src/config/validate-config.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+function isDependencyInstalled(moduleName: string): boolean {
+  try {
+    require.resolve(moduleName);
+    return true;
+  } catch {
+    return false;
+  }
+}
```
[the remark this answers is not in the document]
samarpan-b: Refactored the module resolution check into a shared helper so validation and instrumentation loading use the same dependency lookup path.

packages/observability/src/config/resolve-config.ts
```diff
@@ -0,0 +1,197 @@
...
+  const environment = bootstrapOverrides?.environment ?? env.NODE_ENV;
+
+  const exporterProtocol = (bootstrapOverrides?.exporterProtocol ??
+    (env.OTEL_EXPORTER_OTLP_PROTOCOL as ExporterProtocol | undefined) ??
+    DEFAULT_OBSERVABILITY_CONFIG.exporterProtocol) as ExporterProtocol;
+
+  const sampler = (bootstrapOverrides?.sampler ??
+    (env.OTEL_TRACES_SAMPLER as SamplerName | undefined) ??
+    DEFAULT_OBSERVABILITY_CONFIG.sampler) as SamplerName;
+
```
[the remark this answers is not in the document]
samarpan-b: The invalid-env-value concern is addressed in the latest config refactor. I also cleaned up the file formatting so lint is green again.

packages/observability/src/profiles/otlp.profile.ts
```diff
@@ -0,0 +1,141 @@
...
+      url: config.otlpEndpoint,
+      metadata: Object.keys(config.otlpHeaders).length
+        ? config.otlpHeaders
+        : undefined,
+    });
+  }
+
+  const otlpHttp = require('@opentelemetry/exporter-trace-otlp-http') as {
+    OTLPTraceExporter: new (options?: object) => SpanExporter;
+  };
```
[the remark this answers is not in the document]
samarpan-b: fixed.

packages/observability/src/component.ts
samarpan-b: I agree this is a reasonable follow-up, but I don’t want to expand the first observability package PR into framework-specific sequence behavior. HTTP/Express request spans are already covered by the current OTEL instrumentation path; sequence-level request enrichment and logger correlation can be added separately.

packages/observability/src/types.ts
samarpan-b: This PR is intentionally traces-first. I agree `MeterProvider` support belongs in the package long term, but I’d prefer to add metrics in a follow-up PR rather than expand the initial runtime contract in the first release.

packages/observability/src/runtime.ts
samarpan-b: The singleton is intentionally process-scoped for one bootstrap per process. Moving it to `globalThis` would not materially change that behavior here, and test isolation is already handled via explicit runtime reset helpers plus unit coverage.

packages/observability/src/bootstrap.ts
samarpan-b: I don’t think an additional mutex is necessary in the current implementation. `bootstrapObservability()` is synchronous and has no async gap before runtime installation, so two callers cannot interleave on the JS event loop in the way described here.

packages/observability/src/profiles/otlp.profile.ts
samarpan-b: Addressed in the latest changes. The OTLP exporter loading path now uses typed module imports, and the gRPC path builds proper `Metadata` instead of passing a plain header object.

packages/observability/src/profiles/otlp.profile.ts
samarpan-b: Addressed in the latest changes. `exporterName` now reflects the actual exporter instance rather than the internal OTLP loader helper name.

packages/observability/src/component.ts
samarpan-b: Fixed in the latest commit. The component no longer implicitly bootstraps observability during DI construction. Bootstrap remains an explicit startup concern, and the component now expects observability to be initialized before mounting.

packages/observability/src/correlation.ts
samarpan-b: Fixed in the latest commit. `getTraceContext()` now returns empty correlation fields without creating propagation headers when there is no active span.

packages/observability/src/profiles/instrumentations.ts
samarpan-b: I considered this, but I’m intentionally keeping the explicit instrumentation model in this package. The current design is optimized for consumer-controlled installation, explicit enablement, and clear validation errors for missing optional peers. `@opentelemetry/auto-instrumentations-node` is a valid convenience-oriented alternative, but it shifts the package away from that explicit contract, so I’m not planning to adopt it in this PR.

sf-sahil-jassal (approved):

rohit-sourcefuse (approved):

yeshamavani (approved):

PR #2554, opened 2026-05-28
feat(cache): add sequelize support

### LINKED ISSUE #2552: add Sequelize support
Description:

Added Sequelize integration support in the cache package

Sourav-kashyap:
GH-2552

## Description

Added Sequelize support to the cache package

## Type of change

## Checklist:

### DIFF
--- a/packages/cache/README.md
+++ b/packages/cache/README.md
@@ -51,6 +51,49 @@ export class TestWithMixinRepository extends CacheMixin(
 }
 ```
 
+### Using with Sequelize Repositories
+
+For repositories based on `SequelizeCrudRepository` from `@loopback/sequelize`, use the `SequelizeCacheMixin`:
+
+```ts
+import {SequelizeCacheMixin} from '@sourceloop/cache/sequelize';
+import {SequelizeCrudRepository, SequelizeDataSource} from '@loopback/sequelize';
+
+export class UserRepository extends SequelizeCacheMixin(
+  SequelizeCrudRepository<User, number, {}>,
+  {
+    ttl: 1800,                    // Time to live in seconds (optional)
+    invalidationTags: ['users'],   // Tags for cache invalidation on writes
+    cachedItemTags: ['user-item'], // Tags for individual cache entries
+    disableCachedFetch: false,     // Disable caching if needed (optional)
+  },
+) {
+  cacheIdentifier = 'userRepo';   // Required: unique cache namespace
+
+  constructor(
+    @inject('datasources.postgres') dataSource: SequelizeDataSource,
+  ) {
+    super(User, dataSource);
+  }
+}
+```
+
+**Note:** Import from `@sourceloop/cache/sequelize` for Sequelize-specific caching.
+
+**Difference from CacheMixin:**
+- `CacheMixin` works with `DefaultCrudRepository` (Juggler-based)
+- `SequelizeCacheMixin` works with `SequelizeCrudRepository` (Sequelize-based)
+- Use the appropriate mixin based on your repository type
+
+**Configuration Options:**
+
+| Option | Type | Default | Description |
+|--------|------|---------|-------------|
+| `ttl` | `number` | `86400` (1 day) | Cache expiration time in seconds |
+| `invalidationTags` | `string[]` | `[]` | Tags to invalidate on create/update/delete |
+| `cachedItemTags` | `string[]` | `[]` | Tags to apply to cached items |
+| `disableCachedFetch` | `boolean` | `false` | Bypass cache for all reads |
+
 ### In a controller or service
 
 To add caching to a service or controller, just implement the `ICachedService` interface, adding a binding for the `ICacheService` and the applying the relevant decorators to the methods you want cached -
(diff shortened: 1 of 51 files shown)


sandbox/auth-multitenant-example/src/application.ts
```diff
@@ -80,7 +80,12 @@ export class AuthMultitenantExampleApplication extends BootMixin(
       keyGenerator: rateLimitKeyGen,
     });
     this.bind(HelmetSecurityBindings.CONFIG).to({
-      frameguard: {action: process.env.X_FRAME_OPTIONS},
+      frameguard: {
+        action: process.env.X_FRAME_OPTIONS as
```
rashisf, line 84: is this part of the cache support that you added

Sourav-kashyap: no this changes wasn't part of the cache support. this changes for TypeScript/ESLint type-safety issue where `process.env.X_FRAME_OPTIONS` is typed as `string | undefined`, but `frameguard.action` expects `'deny' | 'sameorigin' | undefined`.

rashisf (approved):

packages/cache/README.md
```diff
@@ -51,6 +51,49 @@ export class TestWithMixinRepository extends CacheMixin(
...
+**Difference from CacheMixin:**
+- `CacheMixin` works with `DefaultCrudRepository` (Juggler-based)
+- `SequelizeCacheMixin` works with `SequelizeCrudRepository` (Sequelize-based)
+- Use the appropriate mixin based on your repository type
+
+**Configuration Options:**
+
+| Option | Type | Default | Description |
+|--------|------|---------|-------------|
+| `ttl` | `number` | `86400` (1 day) | Cache expiration time in seconds |
```
yeshamavani: The docs here describe `ttl` as "Cache expiration time in seconds" with a default of `86400` (1 day), but the actual unit is milliseconds. `DEFAULT_CACHE_OPTIONS.ttl` is `1000 * 60 * 60 * 24 = 86400000`, and that value flows unconverted into the stores: `InMemoryStoreStrategy` compares against `Date.now() + ttl` (ms) and `RedisStoreStrategy` issues `SET ... PX ttl` (PX is milliseconds).

So the documented default is off by 1000x and the unit is wrong. A user copying `ttl: 1800` expecting 30 minutes would actually cache for 1.8 seconds. Please document ttl in milliseconds and fix the default:

```
| `ttl` | `number` | `86400000` (1 day) | Cache expiration time in milliseconds |
```

Same fix applies to the "seconds" comments in the code examples above (lines ~65, 159, 183) and the interface reference (line ~272).

packages/cache/src/mixins/sequelize/cache.mixin.ts
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    @cacheInvalidator(cacheOptions?.invalidationTags)
+    async deleteAll(where?: Where<M>, options?: AnyObject): Promise<Count> {
+      return super.deleteAll(where, options);
+    }
+
+    @cacheInvalidator(cacheOptions?.invalidationTags)
+    async deleteById(id: ID, options?: AnyObject): Promise<void> {
+      return super.deleteById(id, options);
```
yeshamavani, line 207: This whole mixin is a near-verbatim copy of `CacheMixin` in `../cache.mixin.ts` — the five cached read methods, all eleven `@cacheInvalidator`-decorated write methods, and this private `addTagsToOptions` helper are identical. The only real difference is the base type (`SequelizeCrudRepository` vs `DefaultCrudRepository`).

This doubles the maintenance surface: any future caching fix has to be applied in both places or they silently drift (the tag-ordering inconsistency below already shows this starting). Consider extracting the shared body into a single generic mixin factory parameterized over the base repository constructor, or at minimum hoisting `addTagsToOptions` into a shared module imported by both mixins.

packages/cache/src/mixins/sequelize/cache.mixin.ts
```diff
@@ -0,0 +1,223 @@
...
+        options,
+        cacheOptions,
+      );
+    }
+
+    async findOne(
+      filter?: Filter<M>,
+      options?: ICachedMethodOptions,
+    ): Promise<(M & Relations) | null> {
+      options = addTagsToOptions(options, cacheOptions?.cachedItemTags);
```
yeshamavani: `find` and `findOne` call `addTagsToOptions(...)` before the `disableCachedFetch` early-return, but `findById`, `count`, and `exists` call it after. This is behavior-neutral today (the underlying `super` methods ignore `tags`), but the inconsistency is easy to trip over later. Pick one ordering — checking `disableCachedFetch` first is cheapest — and apply it uniformly across all read methods.

packages/cache/package.json
```diff
@@ -45,7 +59,13 @@
     "!*/__tests__"
   ],
   "peerDependencies": {
-    "@loopback/core": "^7.0.3"
+    "@loopback/core": "^7.0.3",
+    "@loopback/sequelize": "^0.8.0"
```
yeshamavani, line 63: `@loopback/sequelize` is correctly declared as an optional peer dependency, but it is imported directly in this package's own source (`src/mixins/sequelize/cache.mixin.ts`), so the package must be able to compile it itself. It is missing from `devDependencies` — the build currently only succeeds because `@loopback/sequelize@0.8.9` is hoisted into the root `node_modules` by sibling packages. If those siblings drop or bump it, this build breaks with no local guarantee.

Please add `@loopback/sequelize` to `devDependencies` (matching the `^0.8.0` peer range) so the package builds in isolation.

yeshamavani (changes requested):
Thanks for adding the Sequelize path — the optional peer dependency + `peerDependenciesMeta`, the subpath export keeping it out of the main barrel, and `ICachedSequelizeRepository` using `FilterExcludingWhere` for `findById` are all done right. A few things I'd like addressed before merge.

### Must fix before merge

- **README `ttl` unit is wrong.** It's documented as seconds with a default of `86400`, but the actual unit is milliseconds (default `86400000`). Off by 1000x for a new public API — details inline.
- **No test coverage for `SequelizeCacheMixin`.** The Juggler `CacheMixin` has integration + acceptance suites with fixture repos; the new mixin and the `@sourceloop/cache/sequelize` entry point have none. Nothing verifies cache hit/miss, invalidation on writes, `disableCachedFetch`, or that the subpath actually resolves. A SQLite/in-memory Sequelize datasource fixture is enough to mirror the existing tests.
- **Unrelated formatting churn.** ~40 files outside `packages/cache` are reformatted (services/*/migration.js, several types.ts, core has-many-resolver) — Prettier-only changes that suggest this branch runs a different Prettier version than master. Please revert these or split them into a separate chore PR so this one stays scoped to the cache feature.

### Nice to have

- De-duplicate the near-verbatim logic between `CacheMixin` and `SequelizeCacheMixin` (at minimum share `addTagsToOptions`).
- Add `@loopback/sequelize` to `devDependencies` so the package builds in isolation rather than relying on monorepo hoisting.
- Make the tag-application ordering consistent across the read methods.
- Minor: the root `"."` export has no explicit `types` condition, unlike `./sequelize`. Not broken today, but worth making consistent.

packages/cache/README.md
Sourav-kashyap: resolved

packages/cache/src/mixins/sequelize/cache.mixin.ts
Sourav-kashyap: resolved

packages/cache/src/mixins/sequelize/cache.mixin.ts
Sourav-kashyap: resolved

packages/cache/package.json
Sourav-kashyap: resolved

PR #2561, opened 2026-06-10
fix(search-service): fix global search issue

### LINKED ISSUE #2553: In Global Search, users are unable to see search results when searching for names containing . (dot).
**Describe the bug**
In Global Search, users are unable to see search results when searching for names containing . (dot), for example: Deal 10.1.

The issue originates from the _formatAndSanitize method in query.builder.ts, which is used by search.provider.ts.

Current implementation:

_formatAndSanitize(param) {
  return param
    .replace(/[^A-Za-z\s0-9]/g, ' ')
    .split(' ')
    .filter(p => p)
    .map(p => `${p}:*`)
    .join('<->');
}

The current logic replaces all non-alphanumeric characters (including .) with spaces before generating the PostgreSQL full-text search query.

Example:

'Deal 10.1'
=> 'Deal 10 1'
=> ['Deal', '10', '1']
=> 'Deal:*<->10:*<->1:*'

Because of this transformation, 10.1 gets split into 10 and 1, causing search mismatches for values containing dots.

**To Reproduce**
- Go to Global Search
- Search for an entity containing . in its name (e.g. Deal 10.1)
- Observe that no matching results are returned

**Expected behavior**
Search should correctly handle values containing . and return matching results without breaking terms like 10.1 into separate tokens.

**Additional context**
Need to update the sanitization logic in _formatAndSanitize so that meaningful characters like . are preserved during search query generation.

Sourav-kashyap:
fix global search issue
GH-2553

## Description
In Global Search, users are unable to see search results when searching for names containing . (dot), for example: Deal 10.1.
The issue originates from the _formatAndSanitize method in query.builder.ts, which is used by search.provider.ts.

## Type of change

## Checklist:

### DIFF
--- a/services/search-service/src/classes/psql/query.builder.ts
+++ b/services/search-service/src/classes/psql/query.builder.ts
@@ -78,11 +78,17 @@ export class PsqlQueryBuilder<T extends Model> extends SearchQueryBuilder<T> {
   }
 
   _formatAndSanitize(param: string) {
-    return param
-      .replace(/[^A-Za-z\s0-9]/g, ' ')
-      .split(' ')
-      .filter(p => p)
-      .map(p => `${p}:*`)
-      .join('<->');
+    let result = param;
+    while (result.endsWith('.')) {
+      result = result.slice(0, -1);
+    }
+
+    return result
+      .replace(/[!&|<>():'"]/g, ' ') // Remove PostgreSQL tsquery special chars
+      .replace(/[^A-Za-z0-9.\s]/g, ' ') // Keep dots, remove other symbols
+      .split(/\s+/) // Split on any whitespace (not just spaces)
+      .filter(p => p && p !== '.') // Filter empty and standalone dots
+      .map(p => `${p}:*`) // Add prefix match operator
+      .join('<->'); // Join with followed-by operator
   }
 }
--- a/services/search-service/src/__tests__/unit/psql/query.builder.unit.ts
+++ b/services/search-service/src/__tests__/unit/psql/query.builder.unit.ts
@@ -6,10 +6,49 @@ import {buildTestsRunner} from '../runner';
 import {PsqlQueryBuilder} from '../../../classes';
 import {expect} from '@loopback/testlab';
 import {testModelList, testModelListWithIdentifier} from '../..';
+import {AnyObject} from '@loopback/repository';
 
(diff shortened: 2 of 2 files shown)


services/search-service/src/classes/psql/query.builder.ts
```diff
@@ -78,11 +78,17 @@ export class PsqlQueryBuilder<T extends Model> extends SearchQueryBuilder<T> {
...
-      .replace(/[^A-Za-z\s0-9]/g, ' ')
-      .split(' ')
-      .filter(p => p)
-      .map(p => `${p}:*`)
-      .join('<->');
-  }
+    let result = param;
+    while (result.endsWith('.')) {
+      result = result.slice(0, -1);
+    }
```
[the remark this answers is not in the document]
Sourav-kashyap: resolved

vaibhavbhalla2505:
I have checked it, and it's working fine now.

before:
Screenshot 2026-06-10 at 4 30 44 PM

now:
Screenshot 2026-06-10 at 4 31 45 PM

vaibhavbhalla2505 (approved):

services/search-service/src/classes/psql/query.builder.ts
```diff
@@ -78,11 +78,17 @@ export class PsqlQueryBuilder<T extends Model> extends SearchQueryBuilder<T> {
...
+    while (result.endsWith('.')) {
+      result = result.slice(0, -1);
+    }
+
+    return result
+      .replace(/[!&|<>():'"]/g, ' ') // Remove PostgreSQL tsquery special chars
+      .replace(/[^A-Za-z0-9.\s]/g, ' ') // Keep dots, remove other symbols
+      .split(/\s+/) // Split on any whitespace (not just spaces)
+      .filter(p => p && p !== '.') // Filter empty and standalone dots
+      .map(p => `${p}:*`) // Add prefix match operator
```
yeshamavani, line 91: does this need a fix in mysql builder as well
check and confirm

Sourav-kashyap: This issue is specific to PostgreSQL's tsquery syntax requirements. MySQL's full-text search handles dots correctly by default.

yeshamavani (approved):

PR #2611, opened 2026-09-14
fix(authentication-service): match login email case insensitively

### LINKED ISSUE #2609: feat(authentication-service): case-insensitive email lookup in login verify providers

Yes. In `KeycloakVerifyProvider` (`services/authentication-service/src/modules/auth/providers/keycloak-verify.provider.ts`), the existing user is resolved with a case-sensitive equality match:

```ts
let user: IAuthUser | null = await this.userRepository.findOne({
  where: {
    email: profile.email,
  },
});
```

Email addresses are case-insensitive in practice (the domain part per RFC, and the local part as treated by virtually all providers — Gmail, Outlook, corporate IdPs). However, identity providers do not guarantee a canonical casing: Keycloak can return `John.Doe@example.com` while the user row was created as `john.doe@example.com` (e.g. via admin user creation, bulk import, or a different signup path).

When the casing differs, `findOne` returns `null`, so a valid existing user is either:
- rejected with `401 InvalidCredentials` (when no signup provider creates users on the fly), or
- **JIT-provisioned as a duplicate user account** by the signup provider, which is worse — the user silently loses access to their existing data.

The same case-sensitive email lookup pattern exists in the other verify/signup flows of the service (e.g. SAML verify provider, local login by username/email, forgot-password), so the issue is not limited to Keycloak.

Make the email match case-insensitive in the login verify providers. Possible approaches:

1. **Normalize at query time** — look up with a case-insensitive operator, e.g. `where: {email: {ilike: <escaped profile.email>}}` (with `%`/`_`/`\` escaped since ILIKE is a pattern match), or lowercase both sides.
2. **Normalize at write time (preferred long-term)** — store `email` lowercased in `users`/`user_credentials` on every create/update path, lowercase the incoming `profile.email` before lookup, and keep plain (index-friendly) equality. Optionally enforce with a unique functional index on `LOWER(email)`.

Either approach could be gated behind an opt-in service config flag (e.g. `emailCaseInsensitive: true` on `AuthServiceBindings.CONFIG`) to avoid changing behavior for existing consumers, since tenants with mixed-case duplicate emails already in the DB would need to resolve collisions first.

**Describe alternatives you've considered**

- Overriding `Strategies.Passport.KEYCLOAK_VERIFIER` in the consuming application with a copy of the stock provider that uses `ilike` — works, but duplicates ~60 lines of library code that drifts on every upgrade.
- Doing a case-insensitive fallback lookup inside the `KEYCLOAK_PRE_VERIFY_PROVIDER` hook (the stock verifier passes the found-or-null user to it and uses its return value) — this is our current workaround, but it relies on a hook whose purpose is verification, not user resolution, and every consumer has to rediscover and re-implement it.

**Additional context**

Suggested acceptance criteria:
- A user whose stored email differs from the IdP profile email only by case can log in via Keycloak and is matched to their existing account (no duplicate JIT-provisioned user).
- Behavior is consistent across Keycloak, SAML, and local login flows.
- Lookup remains SQL-injection safe (parameterized) and documented regarding index usage.

piyushsinghgaur1:
## Problem

Identity providers do not guarantee a canonical casing for email addresses. Keycloak can return `John.Doe@example.com` for a user whose row was created as `john.doe@example.com` — via admin user creation, bulk import, or a different signup path.

The login verify providers resolved the existing user with a case-sensitive equality match:

```ts
let user: IAuthUser | null = await this.userRepository.findOne({
  where: {
    email: profile.email,
  },
});
```

When the casing differed, `findOne` returned `null` and a valid existing user was either:

- rejected with `401 InvalidCredentials`, or
- provisioned as a **duplicate account** by the signup provider, silently losing access to their existing data.

## What this does

**Keycloak and SAML verify providers** now retry the lookup with the lowercased address when the exact match misses:

```ts
const email = profile.email;
const lowerCasedEmail = email.toLowerCase();
let user: IAuthUser | null = await this.userRepository.findOne({
  where: {
    email,
  },
});
// identity providers do not guarantee the casing of the email they
// return, so retry with the canonical form before giving up
if (!user && lowerCasedEmail !== email) {
  user = await this.userRepository.findOne({
    where: {
      email: lowerCasedEmail,
    },
  });
}
```

**`UserRepository`** stores the email in its canonical form, in both the juggler and Sequelize implementations, so rows created through the service are lowercased from the start:

```ts
async create(entity: DataObject<User>, options?: Options): Promise<User> {
  if (entity.email) {
    entity.email = entity.email.toLowerCase();
  }
  ...
}
```

The same normalization is applied in `createWithoutPassword()`. This matches the convention already used for `username`, which `UserRepository.verifyPassword()` has always looked up as `username.toLowerCase()`.

A user whose stored email differs from the IdP profile email only by case can now log in and is matched to their existing account, with no duplicate user created.

## Backward compatibility

This is designed to change the outcome only for logins that fail today.

- **The exact match runs first and is unchanged.** Every login that works today keeps working, byte for byte — including rows deliberately stored with a mixed casing, which still resolve on the first query.
- **The fallback is additive.** It only runs when the first query misses, i.e. on the path that would otherwise have returned `401 InvalidCredentials` or created a duplicate.
- **No extra cost on the normal path.** When the incoming address is already lowercase, `lowerCasedEmail !== email` is false and no second query is issued at all.
- **Deterministic.** Because the exact match is tried first, a user always resolves to the row matching exactly what the identity provider sent.
- **Plain equality, not `ilike`.** Both queries use a parameterized equality match, so the lookup stays index-friendly and portable across every connector this service ships migrations for.
- **No data migration and no configuration required.** Write-side normalization applies only to rows created from here on; existing rows are untouched.
- **No API change.** No constructor signatures, bindings or exported types were modified, so consumers subclassing or manually instantiating these providers are unaffected.

## Testing

- ✅ `lb-tsc` compiles clean
- ✅ `npm test` — **168 passing**
- ✅ `npm run lint` — clean (prettier + eslint)
- ✅ The existing `keycloak-verify.provider.unit.ts` and `saml-verify.provider.unit.ts` suites pass **unmodified**, including their `sinon.assert.calledOnce(findOne)` assertions — confirming the normal path still issues a single query and behaves exactly as before.

Keycloak login scenarios:

| Stored `users.email` | IdP sends | Result |
| --- | --- | --- |
| `john.doe@example.com` | `John.Doe@example.com` | ✅ resolves to the existing user, no duplicate created |
| `john.doe@example.com` | `john.doe@example.com` | ✅ resolves on the first query, unchanged |
| `John.Doe@example.com` | `John.Doe@example.com` | ✅ resolves on the first query, unchanged |

Refs GH-2609


services/authentication-service/src/modules/auth/providers/keycloak-verify.provider.ts
```diff
@@ -40,11 +40,22 @@ export class KeycloakVerifyProvider implements Provider<VerifyFunction.KeycloakA
...
+      const lowerCasedEmail = email.toLowerCase();
       let user: IAuthUser | null = await this.userRepository.findOne({
         where: {
-          email: profile.email,
+          email,
         },
       });
+      // identity providers do not guarantee the casing of the email they
+      // return, so retry with the canonical form before giving up
+      if (!user && lowerCasedEmail !== email) {
```
Tyagi-Sunny, line 52: Same case-sensitivity bug exists in `auth0-verify.provider.ts`, `apple-oauth2-verify.provider.ts`, `cognito-oauth2-verify.provider.ts`, `google-oauth2-verify.provider.ts`, `facebook-oauth-verify.provider.ts`, and `instagram-oauth2-verify.provider.ts` — none of them got this fallback. Worth a follow-up (or extending this PR) so Google/Facebook/etc. logins aren't left with the exact bug this PR fixes for Keycloak/SAML.

services/authentication-service/src/repositories/user.repository.ts
```diff
@@ -107,6 +107,9 @@ export class UserRepository extends DefaultSoftCrudRepository<
   }
 
   async create(entity: DataObject<User>, options?: Options): Promise<User> {
+    if (entity.email) {
```
Tyagi-Sunny, line 110: Minor: this mutates the caller-owned `entity` object in place rather than working on a copy. Low risk given how LoopBack typically passes these in, but worth a note.

Bigger gap: `updateById`/`patch`/`replaceById` on this repository aren't overridden, so an email changed through those paths can be stored non-lowercase, undermining the "rows are canonical from here on" guarantee described in the PR. Same applies to the Sequelize `UserRepository`.

services/authentication-service/src/repositories/sequelize/user.repository.ts
```diff
@@ -101,6 +101,9 @@ export class UserRepository extends SequelizeSoftCrudRepository<
   }
 
   async create(entity: DataObject<User>, options?: Options): Promise<User> {
+    if (entity.email) {
```
Tyagi-Sunny, line 104: Same mutation note as the juggler `UserRepository` — see that file's comment. Keeping both overrides in sync is good, just flagging the same caller-mutation + missing-update-path issue applies here too.

Tyagi-Sunny (commented):
**Solid fix for Keycloak/SAML, but the problem statement is broader than the patch.**

This correctly fixes the case-sensitivity bug for Keycloak and SAML logins, and the fallback query is well-designed (exact match first, no extra query on the common path, existing `calledOnce(findOne)` assertions still hold).

Two things worth resolving before/after merge:

1. **Same bug, other providers untouched.** The PR description frames this as a general "identity providers don't guarantee canonical casing" problem, but `auth0-verify.provider.ts`, `apple-oauth2-verify.provider.ts`, `cognito-oauth2-verify.provider.ts`, `google-oauth2-verify.provider.ts`, `facebook-oauth-verify.provider.ts`, and `instagram-oauth2-verify.provider.ts` all do the identical `findOne({where: {email: profile...email}})` lookup and remain exposed to the same 401/duplicate-account bug. Is this intentionally scoped to Keycloak/SAML only, or should the fallback be extracted into a shared helper and applied to the rest in this PR (or a fast-follow)?
2. **Update paths bypass normalization.** `UserRepository`/Sequelize `UserRepository` only lowercase `email` in `create()`/`createWithoutPassword()`. `updateById`/`patch`/`replaceById` are inherited unchanged, so an email changed through a downstream update endpoint can land mixed-case. If that row's canonical casing later comes in lowercase from the IdP, the new fallback won't catch it either — the retry only fires when the *incoming* profile email differs from its own lowercased form, not when it differs from the stored value. Given this repo is a library other services build CRUD endpoints on top of, this seems like a real (if narrower) gap in the "future writes are canonical" claim in the PR description.

services/authentication-service/src/repositories/user.repository.ts
piyushsinghgaur1: Both addressed.

**Mutation:** `create()` and `createWithoutPassword()` now normalize onto a copy — `entity = {...entity, email: entity.email.toLowerCase()}` — so the caller's payload is no longer rewritten.

**Update paths:** `updateById()` and `replaceById()` are now overridden with the same normalization, each delegating to `super` so the base body/id validation and not-found handling are unchanged. `patch` is the REST operation over `updateById`, so it is covered by that override.

One caveat worth naming: `updateById` is itself a wrapper that delegates to `updateAll`, so a consumer calling `userRepository.updateAll({email}, where)` directly is still uncovered. Happy to extend it there if you think that path is worth guarding.

services/authentication-service/src/repositories/sequelize/user.repository.ts
piyushsinghgaur1: Same changes applied symmetrically here — copy-based normalization in `create()`/`createWithoutPassword()`, plus the `updateById()`/`replaceById()` overrides. Both repositories stay in sync.

services/authentication-service/src/modules/auth/providers/keycloak-verify.provider.ts
piyushsinghgaur1: Not in the scope as of Now Discussed with @a-ganguly Sir

Tyagi-Sunny (approved):
Approving — the mutation and update-path normalization gaps from my earlier review are fixed and verified in the latest commit (852c8fe). The other-IdP-providers coverage gap (thread left open) is a deliberate scope decision per the discussion noted on that thread, not a blocker for this PR.

PR #2613, opened 2026-09-22
feat(benchmarking): add @sourceloop/benchmarking package

sf-sahil-jassal:
## Description

Adds `@sourceloop/benchmarking`, a package that lets a team write throughput benchmarks as Mocha tests and fail the build when a change makes an operation slower.

`bench.describe()` and `bench.it()` wrap tinybench. Each run records its throughput mean in a committed JSON file. The gate compares the current run against the average of the last ten runs and fails the test when the drop is larger than the threshold.

Two rules differ from what bencher, github-action-benchmark and CodSpeed do, and the README says so plainly:

- A regressed run is not added to the history. The window holds until the regression is fixed or accepted, so three merges that each lose 15 percent cannot walk the baseline down past a 40 percent gate one step at a time.
- A new benchmark records samples quietly until it has enough of them. Comparing against a single prior run is a coin flip, not a signal.

The package is ported from an internal `perf-testing` package that is already in use. Behaviour matches it, except for the renamed environment variables (`PERF_*` to `BENCH_*`), the renamed helpers (`pit`/`pdescribe` to `bench.it`/`bench.describe`), the new minimum sample count, and stricter parsing of numeric settings.

Documentation lives in `packages/benchmarking/README.md`. Besides the settings table and the rules the gate applies, it covers how to write a benchmark that means something: good and bad examples side by side, why a `Promise.all` body reports batches per second and not requests per second, and where the noise in a measurement comes from.

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [x] 64 unit tests over the config getters, the reporter, the metric conversion and the Mocha helper
- [x] An acceptance benchmark that runs the package against itself end to end and asserts the sample it recorded
- [x] `npm run build`, `npm run test` and `npm run lint` in the package, all clean

Reproduce with:

```sh
npm run build --workspace @sourceloop/benchmarking
npm run test --workspace @sourceloop/benchmarking
npm run lint --workspace @sourceloop/benchmarking
CI=true npm run test:benchmark --workspace @sourceloop/benchmarking
```

Note that `engines.node` is `^22.12.0 || >=24` for this package, not the usual `22 || 24`. tinybench 5 is ESM only, and `require(esm)` is stable from Node 22.12 onwards. The README explains this.

## Checklist:

- [x] Any dependent changes have been merged and published in downstream modules


packages/benchmarking/README.md
```diff
@@ -0,0 +1,283 @@
...
+
+Benchmark helpers for Mocha. We write a benchmark the same way we write a test, and the build fails when throughput drops further than we allow against a tracked baseline.
+
+The package wraps [tinybench](https://github.com/tinylibs/tinybench) and keeps a rolling history of results in a JSON file that we commit. Nothing else to run, no separate benchmark runner, no dashboard.
+
+## What this is, and what it is not
+
+This is a regression gate on the latency of one logical operation, run serially.
+
+It is not a load test. It does not open connections in parallel, it does not model arrival rates, and it will not tell us how the service behaves at 500 concurrent users. If that is the question, reach for k6 or Artillery. What this answers is narrower and still useful: did the code we just wrote make this operation slower than it used to be.
```
akshatdubeysf: it can open connections in parallel if the code under test does that, specify that anything within the test would be run in parallel, irrespective of what the code actually does.

akshatdubeysf (changes requested):

packages/benchmarking/README.md
sf-sahil-jassal: Good catch, fixed. The line claimed the harness prevents parallelism, which is not true, it only runs iterations serially; whatever the callback does inside one iteration is up to the code under test. Reworded to say that, with a pointer to the Promise.all batches-per-second note below for exactly this case. See 6fafe2ba.

PR #2614, opened 2026-09-23
feat(cli): add code, traced and legacy docker builds to generators

sf-sahil-jassal:
## Description

This PR adds a faster Docker build flow to the monorepos that `sl scaffold` and `sl microservice` generate. Before, each full image ran its own `npm ci` over the whole lockfile. With many services that was slow, and the builds could not run in parallel because the installs used all the agent memory.

Now the pipeline picks one of three modes for each changed service:

- Code build (`Dockerfile.code`): puts the new host-built `dist` on top of the last image for the environment. Most builds use this mode.
- Traced full build (`Dockerfile.nft`, default): starts from one shared deps image (`Dockerfile.deps`, one `npm ci` per pipeline run). Then `scripts/node-file-tracer.js` (`@vercel/nft`) copies only the files that the app loads.
- Legacy full build (`Dockerfile`): the in-image build, kept as a fallback with the `USE_NFT_BUILD` Jenkins parameter.

`scripts/get-full-build-packages.sh` decides which services need a full build. It reads changes to Dockerfiles, dependencies, the `nft` config, patches and the lockfile. Its header comment lists the conventions it depends on, e.g. `@local/<folder>` package names.

### Changes

- Scaffold: `Dockerfile.deps`, `.dockerignore`, the tracer and the change detection script, plus a `docker:build:deps` script and a root `nft` config. The GitHub build workflow now uses the traced build.
- Microservice generator:
  - new `Dockerfile.nft` and `Dockerfile.code`, and a rewritten `Dockerfile`, all with build args and the same runtime layout
  - new scripts: `docker:build:full|nft|code`, `docker:push`, `docker:tag`, `docker:retag`, `helm-update`
  - `nft.alwaysCopy` for each `loopback-connector-*` dependency
  - removed the old `docker:build`, `docker:build:dev` and `docker:push:dev` scripts, and the unused `_appendDockerScript`
- Bug fix: the old scripts passed the base service name as `SERVICE_NAME`. The scripts now use the service folder name.
- Jenkinsfile template:
  - change detection and the code, traced or legacy build split
  - `BUILD_ALL_PACKAGES`, and image promotion with retag
  - Trivy in place of Snyk
  - the jq and yq downloads pick the agent architecture
- Docs in `packages/cli/src/generators/microservice/docs/docker-builds.md`: files, change detection, Jenkins parameters, code builds, traced builds (deps image, what a tracer cannot see, `nft` config, testing), legacy builds, local packages and adoption.
- Sonar and nyc now exclude the new templates.

## Type of change

This changes only files in newly generated projects. Services generated before this change keep their old scripts.

## How Has This Been Tested?

- [x] `npm run build` and `npm test` in `packages/cli` (22 passing)
- [x] Rendered all changed templates with EJS. The rendered `package.json.tpl` parses as JSON.
- [x] Ran `get-full-build-packages.sh` in a test git repo for 10 cases: first build, no change, source only, version bump, local package deps, `Dockerfile.nft` change, `Dockerfile.code` change, lockfile only, root `nft` change, unknown commit. Each case gave the expected output.
- [x] Ran `docker buildx build --check` on all Dockerfile templates.
- [ ] Not done yet: a real traced build and boot of a newly scaffolded project.

## Checklist:


packages/cli/src/generators/microservice/templates/Dockerfile.code
```diff
@@ -0,0 +1,40 @@
...
+#   IMAGE_NAME    image name, i.e. the npm package name
+#   BUILD_ENV     environment tag of the base image, e.g. dev
+#
+# LOCAL PACKAGES: for each `@local/<name>` in this service's dependencies,
+# uncomment and fill in the marked lines, so that their new dist goes in too.
+ARG IMAGE_REPO
+ARG IMAGE_NAME
+ARG BUILD_ENV=dev
+
+FROM ${IMAGE_REPO}/${IMAGE_NAME}:${BUILD_ENV}
```
rohit-sourcefuse, line 21: The container fails to start after a normal code-only change.

When the base image came from `Dockerfile.nft`, its `node_modules` only has the files the old `dist` needed. Say a change starts importing a dependency that's already in package.json but wasn't used before. `get-full-build-packages.sh` returns an empty list (no full-build trigger), so this code build copies the new `dist` on top. In my test repo the image then fails with `Error: Cannot find module 'debug'`. A new deep import like `lodash/fp` would do the same.

What worked for me: trace the new `dist` against the deps image in this build too, and add those files on top of the base. Files already in the base are the same, because a lockfile change forces a full build. With this, the same change boots fine on both an nft base and a legacy base. It also uses the workspace layout from my comment on `Dockerfile`.

```dockerfile
# syntax=docker/dockerfile:1
#
# CODE build. The pipeline uses it when this service changed but no
# full-build trigger fired (see scripts/get-full-build-packages.sh). It
# starts FROM the last image of this service for the environment and puts
# the new host-built dist on top. It does not run npm ci.
#
# A traced base image only has the node_modules files that the OLD dist
# needed. The new dist can import something the old one did not (a declared
# but unused dependency, a new deep import), so we trace the new dist against
# the shared deps image and add its files on top. Files already in the base
# are the same, because a lockfile change forces a full build.
#
# Build context: the repo root. Build args:
#   SERVICE_NAME  folder name of this service, e.g. auth-service
#   FROM_FOLDER   services or facades
#   IMAGE_REPO    registry repo, e.g. the DockerHub org
#   IMAGE_NAME    image name, i.e. the npm package name
#   BUILD_ENV     environment tag of the base image, e.g. dev
#   DEPS_IMAGE    tag of the shared deps image, e.g. my-project-deps:local
#
# LOCAL PACKAGES: for each `@local/<name>` in this service's dependencies,
# uncomment and fill in the marked lines, so that their new dist goes in too.
ARG IMAGE_REPO
ARG IMAGE_NAME
ARG BUILD_ENV=dev
ARG DEPS_IMAGE

FROM ${DEPS_IMAGE} AS trace
ARG SERVICE_NAME
ARG FROM_FOLDER

USER node
WORKDIR /home/node/app

COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/dist ./${FROM_FOLDER}/${SERVICE_NAME}/dist
COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/package.json ./${FROM_FOLDER}/${SERVICE_NAME}/package.json
# [local packages 1/4] one line per local package:
# COPY --chown=node packages/<name>/dist ./packages/<name>/dist

# [local packages 2/4] add `--workspace packages/<name>` per local package.
RUN node scripts/node-file-tracer.js ${FROM_FOLDER}/${SERVICE_NAME}/dist \
      --workspace ${FROM_FOLDER}/${SERVICE_NAME} \
      > /tmp/nft_list.txt \
 && mkdir -p /home/node/pruned/node_modules /home/node/pruned/${FROM_FOLDER}/${SERVICE_NAME} \
 && cpio -pdm /home/node/pruned < /tmp/nft_list.txt

FROM ${IMAGE_REPO}/${IMAGE_NAME}:${BUILD_ENV}
ARG SERVICE_NAME
ARG FROM_FOLDER

WORKDIR /home/node/app

# COPY merges into an existing folder. Without this delete, a file that we
# removed from the source stays in the image, and a booter still loads it.
USER root
RUN rm -rf ./${FROM_FOLDER}/${SERVICE_NAME}/dist
# [local packages 3/4] one line per local package:
# RUN rm -rf ./packages/<name>/dist
USER node

COPY --from=trace --chown=node /home/node/pruned/node_modules ./node_modules
COPY --from=trace --chown=node /home/node/pruned/${FROM_FOLDER}/${SERVICE_NAME}/node_module[s] ./${FROM_FOLDER}/${SERVICE_NAME}/node_modules/
COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/dist ./${FROM_FOLDER}/${SERVICE_NAME}/dist
COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/package.json ./${FROM_FOLDER}/${SERVICE_NAME}/package.json
COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/public ./${FROM_FOLDER}/${SERVICE_NAME}/public
COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/.env.defaults ${FROM_FOLDER}/${SERVICE_NAME}/.env.example ./${FROM_FOLDER}/${SERVICE_NAME}/
# [local packages 4/4] one line per local package:
# COPY --chown=node packages/<name>/dist ./packages/<name>/dist

WORKDIR /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}
```

It needs two small changes around it:

```ts
// microservice/index.ts, _dockerScripts()
'docker:build:code': `${build} --build-arg DEPS_IMAGE=${rootPackageJson.name}-deps:local --build-arg IMAGE_REPO=$IMAGE_REPO_NAME --build-arg IMAGE_NAME=$npm_package_name --build-arg BUILD_ENV=$BUILD_ENV -t ${image} -f Dockerfile.code ../../`,
```

```groovy
// jenkinsfile.tpl, Docker-build: build the deps image for code builds too
// (BuildKit caches it, so it only reinstalls when the lockfile changed).
sh 'echo $DOCKERHUB_CREDENTIALS_PSW | sudo docker login -u $DOCKERHUB_CREDENTIALS_USR --password-stdin'
sh "npm run docker:build:deps"

if (fullBuildPkgs) {
    if (useNft) {
        sh "lerna run docker:build:nft --scope=${scopeFilter} --concurrency 4"
    } else {
        // Each legacy build runs its own npm ci. In parallel they can use all agent memory.
        sh "lerna run docker:build:full --scope=${scopeFilter} --concurrency 1"
    }
    sh "lerna run docker:build:code --ignore=${scopeFilter} ${sinceArg} --concurrency 4"
} else {
    sh "lerna run docker:build:code ${sinceArg} --concurrency 4"
}
```

packages/cli/src/generators/jenkins/templates/jenkinsfile.tpl
```diff
@@ -143,147 +167,174 @@ pipeline {
...
+            steps {
                 script {
+                    def sinceArg = env.BUILD_ALL_PACKAGES.toBoolean() ? "" : "--since ${env.LAST_COMMIT}"
                     sh 'echo $DOCKERHUB_CREDENTIALS_PSW | sudo docker login -u $DOCKERHUB_CREDENTIALS_USR --password-stdin'
-                    if (params.FULL_BUILD){
-                        sh "lerna run docker:push --concurrency 4 -- --IMAGE_REPO_NAME=$IMAGE_REPO_NAME --IMAGE_OPTIONAL_VERSION=$IMAGE_OPTIONAL_VERSION"
-                    } else {
-                        sh "lerna run docker:push  --since $LAST_COMMIT --concurrency 4 -- --IMAGE_REPO_NAME=$IMAGE_REPO_NAME --IMAGE_OPTIONAL_VERSION=$IMAGE_OPTIONAL_VERSION"
-                    }
+                    sh "lerna run docker:push ${sinceArg} --concurrency 4"
```
rohit-sourcefuse, line 287: Images from root-level triggers never get pushed.

When the full build comes from a root file (lockfile, `patches/`, `Dockerfile.deps`, the tracer, root `nft`), Docker-build rebuilds every service. But push, tag, retag and helm-update still run with `--since`, and lerna only diffs inside each package folder. I checked with lerna 9.0.7: after a lockfile-only commit the script returns `other,svc`, while `lerna ls --since` reports `found 0 packages`. So the new images are built and dropped, and `:dev` stays on the old one. That's the exact case these triggers exist for, e.g. a transitive security fix.

A simple fix is to drop `--since` for the rest of the run when the script picked every image. Calling the script with no argument lists all image packages in the same order, so a plain string compare works. In Detect-changes, after the `echo`:

```groovy
                    // A root-level trigger (lockfile, patches/, Dockerfile.deps, tracer, root nft)
                    // marks every image, but `lerna --since` only sees changes inside package
                    // folders, so push, tag and helm-update would skip them. Drop --since then.
                    if (env.FULL_BUILD_PACKAGES && !env.BUILD_ALL_PACKAGES.toBoolean()) {
                        def allImagePackages = sh(
                            script: "bash ${WORKSPACE}/scripts/get-full-build-packages.sh",
                            returnStdout: true
                        ).trim()
                        if (env.FULL_BUILD_PACKAGES == allImagePackages) {
                            env.BUILD_ALL_PACKAGES = 'true'
                        }
                    }
```

packages/cli/src/generators/scaffold/templates/scripts/get-full-build-packages.sh
```diff
@@ -0,0 +1,176 @@
...
+        fi
+        if [[ "$file" == "package.json" ]]; then
+            old_nft=$(git show "$last_commit:package.json" 2>/dev/null | jq -cS '.nft // {}' 2>/dev/null)
+            new_nft=$(jq -cS '.nft // {}' package.json 2>/dev/null)
+            if [[ "$old_nft" != "$new_nft" ]]; then
+                build_all=true
+                break
+            fi
+        fi
+        if [[ "$file" == "package-lock.json" ]]; then
```
rohit-sourcefuse, line 103: Any `*/package.json` in the diff switches this rule off, including a version bump. I tried it with a real npm lockfile: a transitive bump on its own gives `other,svc`, but the same bump plus a version bump in `svc` gives an empty list. `lerna version` bumps package.json on every release, so transitive security fixes will usually never trigger a full build.

Comparing the lockfile entries themselves avoids guessing. The rule then fires when a third-party package that was already in the lockfile resolves to something else. Packages that were only added or removed belong to the workspace whose package.json changed, and rule (d) already covers that. Add this function next to `deps_changed`:

```bash
# Returns 0 when a third-party package that was already in the lockfile now
# resolves to something else (a transitive bump, or a shared hoisted version
# moving). Such a change can reach every image. A package that is only added
# or removed belongs to the workspace whose package.json changed, and rule (d)
# covers that workspace. Workspace version bumps are not node_modules entries,
# so they no longer hide this rule.
lockfile_entry_changed() {
    local changed
    changed=$(jq -n \
        --slurpfile old <(git show "$last_commit:package-lock.json" 2>/dev/null) \
        --slurpfile new package-lock.json '
        def deps: (.packages // {})
            | with_entries(select(.key | test("(^|/)node_modules/"))
                | .value |= {version, resolved, integrity});
        (($old[0] // {}) | deps) as $o | ($new[0] | deps) as $n
        | any($o | keys[]; $n[.] != null and $n[.] != $o[.])') || {
        echo "Error: cannot compare package-lock.json" >&2
        exit 1
    }
    [[ "$changed" == "true" ]]
}
```

and replace this `if` block with:

```bash
        if [[ "$file" == "package-lock.json" ]] && lockfile_entry_changed; then
            build_all=true
            break
        fi
```

Results on the same repo (PR vs this change): transitive bump only `other,svc` / `other,svc`. Transitive bump + version bump `(empty)` / `other,svc`. Version bump only, new dependency in one service, and source-only change are all the same as before. One trade-off: upgrading a hoisted package (say `debug`) now rebuilds every image, even if only one service uses it. I think that's the safer default. After this, the "known gap" note here and in the docs can go.

packages/cli/src/generators/microservice/templates/Dockerfile
```diff
@@ -1,88 +1,61 @@
...
 
-RUN mkdir -p /home/node/app
-
-USER node
-
 WORKDIR /home/node/app
 
+COPY --from=build --chown=node /home/node/app/node_modules ./node_modules
+COPY --from=build --chown=node /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}/dist ./dist
+COPY --from=build --chown=node /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}/node_module[s] ./node_modules/
```
rohit-sourcefuse, line 50: This merges `services/<svc>/node_modules` into the root `node_modules`. npm only nests a package there when its version conflicts with the hoisted one, so after the merge the local packages and other hoisted deps end up with the service's version. With `@local/core` on `ms@2.1.3` and the service pinned to `ms@2.0.0`, both this image and the nft one print `core sees ms@2.0.0`. The old Dockerfile kept the service at `services/<svc>`, so this is new.

Keeping the repo layout in the image fixes it: root `node_modules` at the root, the service at its workspace path with its nested `node_modules`, and `WORKDIR` there. The runtime stage I tested:

```dockerfile
FROM node:${nodeVer}
ARG SERVICE_NAME
ARG FROM_FOLDER

WORKDIR /home/node/app

# Keep the repo layout: root node_modules at the root, the service at its
# workspace path with its own nested node_modules. Merging the nested folder
# into the root one gives every package the service's versions.
COPY --from=build --chown=node /home/node/app/node_modules ./node_modules
COPY --from=build --chown=node /home/node/app/packag[e]s ./packages/
COPY --from=build --chown=node /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}/dist ./${FROM_FOLDER}/${SERVICE_NAME}/dist
COPY --from=build --chown=node /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}/node_module[s] ./${FROM_FOLDER}/${SERVICE_NAME}/node_modules/
COPY --from=build --chown=node /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}/package.json ./${FROM_FOLDER}/${SERVICE_NAME}/package.json
COPY --from=build --chown=node /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}/public ./${FROM_FOLDER}/${SERVICE_NAME}/public
COPY --from=build --chown=node /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}/.env.defaults /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}/.env.example ./${FROM_FOLDER}/${SERVICE_NAME}/

WORKDIR /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}
USER node

ENV HOST=0.0.0.0 PORT=3000
EXPOSE ${PORT}

CMD [ "node", "-r", "./dist/opentelemetry-registry.js", "." ]
```

`Dockerfile.nft` needs the same change:

```dockerfile
FROM node:${nodeVer}
ARG SERVICE_NAME
ARG FROM_FOLDER

WORKDIR /home/node/app

# Same layout as the repo (see Dockerfile). Nested node_modules stay nested.
COPY --from=trace --chown=node /home/node/pruned/node_modules ./node_modules
COPY --from=trace --chown=node /home/node/pruned/${FROM_FOLDER}/${SERVICE_NAME}/node_module[s] ./${FROM_FOLDER}/${SERVICE_NAME}/node_modules/
COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/dist ./${FROM_FOLDER}/${SERVICE_NAME}/dist
COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/package.json ./${FROM_FOLDER}/${SERVICE_NAME}/package.json
COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/public ./${FROM_FOLDER}/${SERVICE_NAME}/public
COPY --chown=node ${FROM_FOLDER}/${SERVICE_NAME}/.env.defaults ${FROM_FOLDER}/${SERVICE_NAME}/.env.example ./${FROM_FOLDER}/${SERVICE_NAME}/
# [local packages 3/3] three lines per local package:
# COPY --chown=node packages/<name>/dist ./packages/<name>/dist
# COPY --from=trace --chown=node /home/node/pruned/packages/<name>/node_module[s] ./packages/<name>/node_modules/
# COPY --chown=node packages/<name>/package.json ./packages/<name>/package.json

WORKDIR /home/node/app/${FROM_FOLDER}/${SERVICE_NAME}
USER node

ENV HOST=0.0.0.0 PORT=3000
EXPOSE ${PORT}

CMD [ "node", "-r", "./dist/opentelemetry-registry.js", "." ]
```

With both (plus the `Dockerfile.code` from my other comment), core gets `2.1.3` and the service keeps `2.0.0` in all three build modes. Existing images have the old layout, so the first build after this should be a full one (`BUILD_ALL_PACKAGES=true`).

packages/cli/src/generators/microservice/index.ts
```diff
@@ -340,22 +340,32 @@ export default class MicroserviceGenerator extends AppGenerator<MicroserviceOpti
...
+    const envImage = '$IMAGE_REPO_NAME/$npm_package_name';
+    const build = `DOCKER_BUILDKIT=1 sudo docker build --build-arg SERVICE_NAME=${serviceName} --build-arg FROM_FOLDER=${folder}`;
+    const helmKey = serviceName.replace(/-([a-z0-9])/g, (_, c: string) =>
+      c.toUpperCase(),
     );
+    return {
+      'docker:build:full': `${build} -t ${image} -f Dockerfile ../../`,
+      'docker:build:nft': `${build} --build-arg DEPS_IMAGE=${rootPackageJson.name}-deps:local -t ${image} -f Dockerfile.nft ../../`,
+      'docker:build:code': `${build} --build-arg IMAGE_REPO=$IMAGE_REPO_NAME --build-arg IMAGE_NAME=$npm_package_name --build-arg BUILD_ENV=$BUILD_ENV -t ${image} -f Dockerfile.code ../../`,
+      'docker:push': `sudo docker push ${image}`,
```
rohit-sourcefuse, line 364: The generated `build-image.yaml` logs in with `docker/login-action`, which writes the runner user's `~/.docker/config.json`. This push runs through `sudo`, and on Ubuntu `sudo` resets `HOME` to `/root`. I checked on `ubuntu:24.04`: user `HOME=/home/runner`, `sudo HOME=/root`. So root has no credentials and the push is denied. The old script ran `docker push` without sudo.

Since the build scripts use `sudo docker` everywhere (and Jenkins logs in with `sudo docker login`), the smallest fix is to log in as root in the workflow as well:

```yaml
      - name: Login to Docker Hub
        # The docker scripts run `sudo docker ...`, and sudo resets HOME to /root,
        # so root needs its own login.
        env:
          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
        run: echo "$DOCKERHUB_TOKEN" | sudo docker login -u "$DOCKERHUB_USERNAME" --password-stdin
```

packages/cli/src/generators/jenkins/templates/jenkinsfile.tpl
```diff
@@ -97,43 +100,64 @@ pipeline {
...
-        stage('Snyk-analysis') {
+        stage('Trivy-analysis') {
             when {
-                expression { params.RUN_SNYK_CHECKS == true }
+                expression { params.RUN_TRIVY_CHECKS }
             }
             steps {
                 script {
-                    sh "if [ \$SNYK_FAIL_BUILD = 1 ]; then sudo npm install snyk -g && snyk auth $SNYK_TOKEN && snyk test --org=ashishkaushik --strict-out-of-sync=false --fail-on=all --all-projects  --policy-path=./.snyk --severity-threshold=$SNYK_SEVERITY_THRESHOLD; fi"
+                    sh "curl -sfL https://raw.githubusercontent.com/aquasecurity/trivy/main/contrib/install.sh | sudo sh -s -- -b /usr/local/bin latest"
```
rohit-sourcefuse, line 135: This pipes Trivy's install script from `main` into `sudo sh` and installs `latest`, on the agent that holds the DockerHub, GitHub and Sonar credentials. The installer only checks the binary against a checksum file from the same release, so a bad release passes. That's what happened in March this year (malicious v0.69.4, CVE-2026-33634). Our `trivy.yaml` workflow template already pins `trivy-action` by SHA for the same reason.

Running the official image pinned by digest avoids installing anything on the agent:

```groovy
                    // Pinned by digest. Bump on purpose: trivy's release channel was compromised in March 2026.
                    sh 'sudo docker run --rm -v "$PWD:/src" -w /src aquasec/trivy:0.74.0@sha256:62b1e65e8869bc4b4c6aa4fa2b21595256c7c2f6018a9d9ad61caf87187c1969 -c trivy.yml fs .'
```

I ran this against the scaffold's `trivy.yml`. It passes on a clean repo and exits 1 on a HIGH finding (lodash 4.17.20, CVE-2021-23337), so the `exit-code: 1` gate still works.

packages/cli/src/generators/jenkins/templates/jenkinsfile.tpl
```diff
@@ -143,147 +167,174 @@ pipeline {
...
+                    withEnv(["SOURCE_ENV=${sourceEnv}"]) {
+                        sh "lerna run docker:retag ${sinceArg} --concurrency 4"
+                    }
                 }
             }
         }
         stage('Docker-build') {
             when {
-                expression { params.BUILD_NEEDED == true }
+                expression { params.BUILD_NEEDED && (env.BUILD_ENV == 'dev' || params.IS_HOTFIX_RELEASE || params.BYPASS_IMAGE_PROMOTION) }
```
rohit-sourcefuse, line 247: Two things around promotion:

1. The docs say the first environment in `PROMOTION_CHAIN` builds images, but this gate checks `'dev'`. With a chain like `develop,qa,production`, `develop` neither retags (idx 0) nor builds, and Docker-push then fails on a missing image. A `BUILD_ENV` that isn't in the chain only prints a warning in Docker-retag.

```groovy
        stage('Docker-build') {
            when {
                expression {
                    def chain = params.PROMOTION_CHAIN.tokenize(',').collect { it.trim() }
                    params.BUILD_NEEDED && (env.BUILD_ENV == chain[0] || params.IS_HOTFIX_RELEASE || params.BYPASS_IMAGE_PROMOTION)
                }
            }
```

and in Docker-retag:

```groovy
                    if (idx == -1) {
                        error("BUILD_ENV=${env.BUILD_ENV} is not in PROMOTION_CHAIN=${params.PROMOTION_CHAIN}")
                    }
```

2. `docker:retag` pulls whatever `:dev` points to right now. If dev has moved on since the commit being promoted, qa/production gets unreleased code under the release version. Recording the commit on the image and checking it before the retag catches that:

```ts
// microservice/index.ts, _dockerScripts()
const build = `DOCKER_BUILDKIT=1 sudo docker build --label org.opencontainers.image.revision=$(git rev-parse HEAD) --build-arg SERVICE_NAME=${serviceName} --build-arg FROM_FOLDER=${folder}`;

'docker:retag': `sudo docker pull ${envImage}:$SOURCE_ENV && rev=$(sudo docker inspect -f '{{ index .Config.Labels "org.opencontainers.image.revision" }}' ${envImage}:$SOURCE_ENV) && { git merge-base --is-ancestor "$rev" HEAD || { echo "${envImage}:$SOURCE_ENV was built from '$rev', which is not in this branch" >&2; exit 1; }; } && sudo docker tag ${envImage}:$SOURCE_ENV ${image}`,
```

I ran both scripts through `npm run` against a local registry. When the qa branch contains the dev commit, the retag goes through. When `:dev` was built from a commit qa doesn't have, it stops with that message and exits 1. Images pushed before this have no label, so they need one dev build before their first promotion.

rohit-sourcefuse (changes requested):
Hi Sahil, nice work on this. 

The shared deps image instead of an `npm ci` per service is a real speedup, and the docs are genuinely useful. Good to see the personal git identity, the Snyk org and the `gobinaries | bash` step gone too.

I spent some time on the change detection and the build flow, and tried it out in a small two-service monorepo built from these templates (with Docker and lerna 9.0.7). A few cases end up shipping stale or broken images while the build stays green, so I'd like these sorted before merge. 

**Details and fixes are inline, and I tested each fix in the same repo:**
1. A code build on top of a traced image can't load a newly imported module.
2. Full builds triggered by root files (lockfile, `patches/`, `Dockerfile.deps`, tracer, root `nft`) are never pushed, tagged or deployed.
3. The lockfile rule is skipped whenever any package.json changes, including a version bump.
4. The service's nested `node_modules` is merged into the root one, so shared packages get the service's versions.
5. `sudo docker push` fails in the generated GitHub workflow.
6. Trivy is installed unpinned, as root.
7. The build gate is hardcoded to `dev`, and promotion can retag a `:dev` built from commits the target branch doesn't have.

**Smaller things, fine as follow-ups:**
- `sudo npm install -g lerna@^9` isn't pinned. lerna is already in the root devDependencies, so `npx lerna` after `npm ci` would do.
- `get-full-build-packages.sh` prints an empty list and exits 0 if jq is missing or `git diff` fails, which means code builds for everything. A `command -v jq` check and `set -e` would catch it.
- With `docker:build` removed, an older Jenkinsfile silently skips newly generated services, and the service README template still mentions it. Maybe keep it as an alias for `docker:build:full`.
- `--helmPath` now needs to start with `<project>-helm/`, otherwise helm-update edits a file outside the helm clone. Worth a line in the docs and the flag help.
- A few tests for the script would be great. It's easy to drive with temp git repos.

